PR #9, opened 2018-12-10
Add Linting and UMD Builds

dbouwman:
In order to use this package in hub, we need a UMD build so I added that. When I first ran it I got a lot of errors, and I noticed the project did not have a `tslint.json` file, so I added that, and run commands to autofix everything it could. 

There are still a few outstanding lint errors - since I'm not deep into this code yet, I did not want to go renaming things wildly. Please take a look though.

In order to get rollup working I had to change the  `tsconfig.json` compileOptions.module to `es2015` from `umd` - looking at the outputs, they *seem* the same as before, and line up with the output from ArcGIS Rest Js - but again, please try this out.

### Update...

I talked with @tomwayson … apparently the default typescript umd output is non-standard in that it does not actually “bundle” the output. So – I’m adding a separate build (`npm run build:tsumd`) and updating the demo/* files to use that output… but from the sounds of it we really should be using the rollup based umd output (`npm run build:umd`)


MikeTschudi:
Thanks, @dbouwman!  I added a commit fixing the remaining lint errors and bringing coverage back up to 100%. test:chrome fine; test:node doesn't work with es2015.

dbouwman:
Thanks @MikeTschudi - I'll pull this and get the node tests running again!

MikeTschudi:
@dbouwman, the arcgis-rest-js node tests don't run for me, either; instead of arcgis-clone-js' problem with `import`, `export` is the problem. Perhaps a configuration error on my part?

PR #12, opened 2018-12-19
F/modularize fullitem

MikeTschudi:


demos/demo_common/polyfills.js
```diff
@@ -0,0 +1,48 @@
+// Source: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/findIndex#Polyfill
```
dbouwman: Not sure if this has to run on-prem w/o internet access, but one solution we have used is to leverage https://polyfill.io/ - whats nice is that the service will only return the needed polyfills for the browser... so if it's chrome that makes the request, it returns an empty script, vs IE9, where it returns a mess of JS to polyfill all the things. You can also specify the things you want - again to minimize what's packaged and sent. Just an option...

src/itemTypes/group.ts
```diff
@@ -0,0 +1,99 @@
...
+        const trancheIds:string[] = contents.items.map((item:any) => item.id);
+
+        // Are there more contents to fetch?
+        if (contents.nextStart > 0) {
+          pagingRequest.paging.start = contents.nextStart;
+          getGroupContentsTranche(id, pagingRequest)
+          .then(
+            (allSubsequentTrancheIds:string[]) => {
+              // Append all of the following tranches to this tranche and return it
+              Array.prototype.push.apply(trancheIds, allSubsequentTrancheIds);
```
dbouwman: Perhaps using .concat could make the intent more clear here? (tbh - I did not know that using Array.prototype.push.apply acted like a .concat, but w/o the reassignment) 
`trancheIds = trancheIds.concat(allSubsequentTrancheIds);`
I think you'd also have to swap the `const trancheIds...` to a `let trancheIds...`

src/itemTypes/webmap.ts
```diff
@@ -0,0 +1,112 @@
...
+ * @return List of ids and/or URLs
+ * @protected
+ */
+export function getWebmapLayerIds (
+  layerList: any
+): string[] {
+  const dependencies:string[] = [];
+
+  if (Array.isArray(layerList)) {
+    layerList.forEach((layer:any) => {
```
dbouwman: You could do this via a .reduce as well...

src/itemTypes/webmappingapplication.ts
```diff
@@ -0,0 +1,73 @@
...
+ */
+export function getDependencies (
+  fullItem: ITemplate,
+  requestOptions: IUserRequestOptions
+): Promise<string[]> {
+  return new Promise(resolve => {
+    const dependencies:string[] = [];
+
+    const values = fullItem.data && fullItem.data.values;
+    if (values) {
```
dbouwman: we have a `getProp` helper fn in hub.js. It's nice in that you can skip these nested guards... and compose more refined helpers like `getProps`, which them allows you to do things like... (from the Hub's current system of getting item dependencies)

```
function getProps (model, props) {
  return props.reduce((a, p) => {
    let v = getProp(model, p);
    if (v) {
      a.push(v);
    }
    return a;
  }, []);
}
export function processGenericWebApp (model) {
  const props = ['data.webmap', 'data.itemId', 'data.values.webmap'];
  return getProps(model, props);
}

```
Here is where the underlying getProp fn definition is - it's sooo small, it's likely better to copy in vs import another pacakge: 
https://github.com/Esri/hub.js/blob/master/packages/common/src/util.ts#L30

test/fullItem.test.ts
```diff
@@ -313,7 +318,7 @@ describe("Module `fullItem`: fetches the item, data, and resources of an AGOL it
 
         const abc = {...MOCK_ITEM_PROTOTYPE};
         abc.type = "Group";
-        abc.item.id = "grp1234567890";
+        abc.item = {id: "grp1234567890"};
```
dbouwman: This seems repeated a number of times... you could streamline this with Object.assign...
`const abc = Object.assign({}, MOCK_ITEM_PROTOTYPE, {type: "Group", item: {id: 'grp1234567890'}});`

demos/demo_common/polyfills.js
MikeTschudi: We do have to run w/o internet access; this is one of the reasons that switching to Hub's templates over references to AGOL source items was hugely useful.

dbouwman: Cool - for our Portal builds we package in the polyfills just like this :)

src/itemTypes/webmap.ts
MikeTschudi: @dbouwman, something like this?  (It passes tests, but probably can be improved.)

```typescript
export function getWebmapLayerIds (
  layerList: any
): string[] {
  return !Array.isArray(layerList) ? [] :
  layerList.reduce(
    (ids:string[], layer:any) => {
      const itemId = layer.itemId as string;
      if (itemId) {
        ids.push(itemId);
      }
      return ids;
    },
    [] as string[]
  );
}
```

src/itemTypes/group.ts
MikeTschudi: How could

```typescript
resolve(trancheIds.concat(allSubsequentTrancheIds));
```

possibly be clearer than

```typescript
Array.prototype.push.apply(trancheIds, allSubsequentTrancheIds);
resolve(trancheIds);
```
;-)

src/itemTypes/webmap.ts
dbouwman: 👍

src/itemTypes/group.ts
dbouwman: Exactly :)

src/itemTypes/webmappingapplication.ts
MikeTschudi: Very nice; thanks!

dbouwman:
@MikeTschudi this looks great! merge away!

PR #17, opened 2019-02-20
 Renamed Solution template functions to match opendata-ui

MikeTschudi:
Would you please do an initial review of the [README](https://github.com/Esri/arcgis-clone-js/tree/chore/renaming) and [background info](https://github.com/Esri/arcgis-clone-js/blob/chore/renaming/docs/Background.md) to see if the big-picture view of the library makes sense?

Because this PR has more changes than branch chore/update-classifier, when the review is done and any changes made, let's wait to merge it until after chore/update-classifier is merged.


jmhauck:
@MikeTschudi sounds good. I started looking through but will continue when I get back into the office in the AM.

test/solution.test.ts
```diff
@@ -509,11 +378,19 @@ describe("Module `solution`: generation, publication, and cloning of a solution
       .post("path:/sharing/rest/content/users/casey/fld1234567890/addItem", addItemUpdater)
       .post("path:/sharing/rest/content/users/casey/items/wma1234567890/update",
         '{"success":true,"id":"wma1234567890"}')
-      .post("path:/sharing/rest/content/users/casey/items/map1234567890/update",
+      .post("path:/sharing/rest/contslent/users/casey/items/map1234567890/update",
```
jmhauck: typeo?

src/itemTypes/classifier.ts
```diff
@@ -113,28 +126,12 @@ export function initItemTemplateFromId (
...
-                    .reduce((acc, val) => acc.concat(val), [])  // some dependencies come out as nested, so flatten
-                  ) as string[]
-                );
-
+              template => {
+                itemTemplate.dependencies = // some dependencies come out as nested, so flatten
+                  removeDuplicates(flatten(template.dependencies));
                 resolve(itemTemplate);
               },
               () => reject({ success: false })
```
jmhauck: Any issues if we allow potential system error to pass through these?

Something like:

(e) => reject({ success: false, error: e })

jmhauck:
After reading through your changes I'm not sure it makes sense to merge chore/update-classifier. The same things it does are captured here in your changes. One difference is that when I looked at classifier I left the item specific call for the dependencies in and just changed how the promises were handled to ensure sequential processing. Should be able to accomplish the same thing in the way that you have approached by moving that item specific function call for the dependencies out of classifier and force it to be called within the item specific code. I think how you did that was better. 

If we don't merge that one we won't have to sort through the conflicts. Also chore/update-classifier does not handle all of the necessary function renames as they really needed to occur across the board for the TS stuff to not start complaining. 

After this is merged I can pull this stuff and work in the other view related code I have been working on. Have really only needed to adjust things in featureservice at this point for view.

jmhauck:
Oh ya...one other thought, that I mentioned on the call...and I know you have been through many changes around naming and structure so I'm not wanting to reopen anything but just had a thought as I was reading through the doc.

Overall I think the doc is good and explains the process well. One thing that I think can be a little confusing is discussing solution template item and solution item templates. I understand that the solution item is essentially a template but wonder if if would help with discussing if template was not used when discussing the solution item. That way we have this item that stores templates rather than a template that stores templates.

test/solution.test.ts
MikeTschudi: Obviously a critical fetchMock  :-/
Removed

src/itemTypes/classifier.ts
MikeTschudi: This seems much more useful. Let's switch to that model for the future and I can retroactively update the older code.

jmhauck: I tried this out in a few places to get more info on an error I am getting with the view and realized that this can lead to some nasty nesting of the actual message we are after.

image

This seems to be a little better:

`(e) => reject({ success: false, error: e.error ? e.error : e })`

image

MikeTschudi: Much better!

MikeTschudi:
@jmhauck, would you please review my updates to [README](https://github.com/Esri/arcgis-clone-js/tree/chore/renaming)'s Common terms and Modules sections and updated to [Background](https://github.com/Esri/arcgis-clone-js/blob/chore/renaming/docs/Background.md)?

Background doesn't have updates to function names and function arguments yet. I was thinking of

#### Creating a solution item
* createSolutionTemplate --> *createSolutionItem*(...): Promise&lt;*ISolutionItem*&gt;
* convertItemToTemplate(itemId, requestOptions, *existingItemTemplates*?)

#### Deploying a solution item
* createSolutionFromTemplate  --> *deploySolutionItem* (*solutionItem*,...)
* createDeployedSolutionItem (title, *solutionItem*,...)
* createItemFromTemplateWhenReady (itemId, *itemTemplates*,...)
* createItemFromTemplate (*itemTemplate*,...)

Do these changes contribute to clarification?

jmhauck:
Looks good Mike. I think these help clarify more.

MikeTschudi:
Thanks; I'll update the code now

PR #20, opened 2019-02-26
Save template thumbnails and resources to an AGOL item

MikeTschudi:
Working on its tests now; wanted to keep PR shorter


src/itemTypes/common.ts
```diff
@@ -214,7 +220,7 @@ export function updateItemData (
       updateResp => {
         resolve(id);
       },
-      () => reject()
+      () => reject({ success: false })
```
jmhauck: Maybe I'm missing it from just seeing this little bit...why is this only rejecting with `{ success: false }`

MikeTschudi: Just hadn't gotten to updating the all of the rejects to the new form

test/mocks/agolItems.ts
```diff
@@ -714,7 +724,13 @@ export function createAGOLRelationship (
 export function getAnImageResponse (
 ): any {
   const fs = require("fs");
-  return fs.createReadStream("./test/mocks/success.png");
+  if (fs.createReadStream) {
+    // Node test
+    return fs.createReadStream("./test/mocks/success.png");
+  } else {
+    // Chrome test
+    return new Blob([atob(imageAsDataUri(false))], {type: "image/png"})
```
jmhauck: minor issue but missing a `;`

MikeTschudi: Thank you; I probably should add linting to the tests

PR #26, opened 2019-04-02
add view support

jmhauck:
basic view support...does not account for field name changes when going to portal


jgravois:
let me know if you could use a hand debugging the failing tests.

jmhauck:
@jgravois I would like chat and ask a few questions when you get a moment

PR #51, opened 2019-06-11
feature service updates

jmhauck:
Fixes:
 templatizing webmap layer references
 visible field mismatch between source view and new view
 feature service name undefined in some cases...will now use name or title


jmhauck:
@MikeTschudi it does need the url but the url is already templatized and will be handled by replaceInTemplate so the line was redundant.

MikeTschudi:
Cool!  Thanks.

Issue #36, opened 2019-06-06
Folder Naming Convention


MikeTschudi:
@chris-fox:
> When creating a new folder with a deployment we should default the folder name to be the same as the Solution, i.e 'Manage Right of Way Activities'. If that folder already exists we should add a 1, 2, 3, etc. continuing to increment with each instance.

> Manage Right of Way Activities
Manage Right of Way Activities 1
Manage Right of Way Activities 2

MikeTschudi:
@chris-fox, if deployment org contains folders
> Manage Right of Way Activities
Manage Right of Way Activities 2

this implementation will create folder "Manage Right of Way Activities 1", the first available name in the series.

chris-fox:
Verified

PR #52, opened 2019-06-11
F/folder naming 36

MikeTschudi:
cf. #36


packages/common/test/restHelpers.test.ts
```diff
@@ -91,12 +1069,13 @@ describe("Module `restHelpers`: common functions involving the arcgis-rest-js li
...
-      restHelpers
-        .createUniqueFolder(folderTitleRoot, MOCK_USER_SESSION)
-        .then((response: portal.IAddFolderResponse) => {
+      createUniqueFolder(folderTitleRoot, MOCK_USER_SESSION).then(
+        (response: portal.IAddFolderResponse) => {
           expect(response).toEqual(expectedSuccess);
           done();
-        }, done.fail);
+        },
+        done.fail
```
jmhauck: I didn't think about doing it like this. I like it...even cleaner than `() => done.fail()`

PR #77, opened 2019-08-05
Handled missing mocks

MikeTschudi:


packages/common/test/restHelpers.test.ts
```diff
@@ -173,6 +173,19 @@ const MOCK_USER_REQOPTS: IUserRequestOptions = {
   authentication: MOCK_USER_SESSION
 };
 
+const SERVER_REST_INFO = {
+  currentVersion: 10.1,
+  fullVersion: "10.1",
+  soapUrl: "https://services123.arcgis.com/org1234567890/arcgis/rest/services",
+  secureSoapUrl:
+    "https://services123.arcgis.com/org1234567890/arcgis/rest/services",
+  owningSystemUrl: "https://myorg.maps..arcgis.com",
```
jmhauck: Not sure if the .. would cause anything unexpected

packages/common/test/resourceHelpers.test.ts
```diff
@@ -97,6 +97,16 @@ const TINY_PNG_BYTES = [
   130
 ];
 
+const SERVER_REST_INFO = {
+  currentVersion: 10.1,
+  fullVersion: "10.1",
+  soapUrl: "https://services123.arcgis.com/org1234567890/arcgis/rest/services",
+  secureSoapUrl:
+    "https://services123.arcgis.com/org1234567890/arcgis/rest/services",
+  owningSystemUrl: "https://myorg.maps..arcgis.com",
```
jmhauck: Noticed the lower one first but the ".." is the only thing that catches my eye.

MikeTschudi:
@jmhauck, thank you for the catch.  I changed both resourceHelpers and restHelpers to use the SERVER_INFO model already in resourceHelpers

PR #82, opened 2019-08-12
Unit test coverage of functions

MikeTschudi:


packages/common/src/resourceHelpers.ts
```diff
@@ -155,6 +155,7 @@ export function copyBlobToStorageItem(
   storageRequestOptions: auth.IUserRequestOptions
 ): Promise<string[]> {
   return new Promise<string[]>((resolve, reject) => {
+    console.warn("========== TODO ==========");
```
jmhauck: Any need for me to mention when these TODOs come through a PR?

MikeTschudi: I don't think so; the goal would simply be to get to 0 "pending specs"/"skipped tests"

packages/common/test/generalHelpers.test.ts
```diff
@@ -361,4 +361,11 @@ describe("Module `generalHelpers`: common utility functions shared across packag
       expect(regEx.test(timestamp)).toBe(true);
     });
   });
+
+  describe("_padPositiveNum", () => {
+    xit("_padPositiveNum", done => {
```
jmhauck: "xit" was a new one for me...had to look it up. Good to know, thanks

Issue #90, opened 2019-08-23
Feature Layer View definition issues


chris-fox:
Discovered a couple issues with the definition of views. To repro deploy 05ad40176eca46b28b6fc99bc55835c1

1. Variables for viewDefinitionQuery and definitionQuery are not being replaced:
"viewDefinitionQuery" : "({{02d58f8ad9fa4cc497f95a319af14b0d.fieldInfos.layer0.fields.funded}} = 'Yes') AND ({{02d58f8ad9fa4cc497f95a319af14b0d.fieldInfos.layer0.fields.projtype}} = 'Water Distribution') AND ({{02d58f8ad9fa4cc497f95a319af14b0d.fieldInfos.layer0.fields.projstatus}} = 'Board Authorized') AND ({{02d58f8ad9fa4cc497f95a319af14b0d.fieldInfos.layer0.fields.projphase}} \u003c\u003e 'Complete')", 
  "definitionQuery" : "({{02d58f8ad9fa4cc497f95a319af14b0d.fieldInfos.layer0.fields.funded}} = 'Yes') AND ({{02d58f8ad9fa4cc497f95a319af14b0d.fieldInfos.layer0.fields.projtype}} = 'Water Distribution') AND ({{02d58f8ad9fa4cc497f95a319af14b0d.fieldInfos.layer0.fields.projstatus}} = 'Board Authorized') AND ({{02d58f8ad9fa4cc497f95a319af14b0d.fieldInfos.layer0.fields.projphase}} \u003c\u003e 'Complete')"

2. It is possible for a view to have a different domain then the source service. In the example see the projtype field. In these scenarios we need to update the definition of the view to pass the correct domain. 

Important, we don't want to do this for every field, we only should do this when the domain is different from the source field. If you update the domain definition for every field, it breaks the connection between the original service domain and the view domain and means if you update the source it won't update on the view.

This is a bug in my current python code because the majority of time the domains will be the same and users will want to update the domain on the source and have it update the view automatically.

chris-fox:
Verified

PR #96, opened 2019-09-04
handle view field domain and defQueries issue

jmhauck:
#90

### DIFF
--- a/packages/feature-layer/src/featureServiceHelpers.ts
+++ b/packages/feature-layer/src/featureServiceHelpers.ts
@@ -100,6 +100,19 @@ export function templatize(
 }
 
 /**
+ * Delete key properties that are system managed
+ *
+ * @param layer The data layer instance with field name references within
+ */
+export function deleteViewProps(layer: any) {
+  const props: string[] = ["definitionQuery"];
+
+  props.forEach(prop => {
+    common.deleteProp(layer, prop);
+  });
+}
+
+/**
  * Cache properties that contain field references
  *
  * @param layer The data layer instance with field name references within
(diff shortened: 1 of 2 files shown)


packages/feature-layer/src/featureServiceHelpers.ts
```diff
@@ -680,6 +705,52 @@ export function _getFieldVisibilityUpdates(fieldInfo: any): any[] {
...
+  // add an update when the domains don't match
+  fieldInfo.newFields.forEach((field: any) => {
+    const i: number = domainNames.indexOf(
+      String(field.name).toLocaleLowerCase()
+    );
+    if (i > -1 && field.hasOwnProperty("domain") && field.domain) {
+      if (JSON.stringify(field.domain) !== JSON.stringify(domainFields[i])) {
+        // should mixin the update if the field already has some other update
+        let hasUpdate: boolean = false;
+        fieldUpdates.forEach((update: any) => {
```
MikeTschudi: How about using [`some`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some) instead of `forEach`? That way, you can break out of the loop after the match is found.

jmhauck: @MikeTschudi I'd not used that before but I like it...thanks! Will switch it up

jmhauck: @MikeTschudi actually I still need to alter the array by setting the domain property on the instance where the name matches. I could still use a standard for loop and break out when the match is found though.

MikeTschudi: The change would be to
```
fieldUpdates.some((update: any) => {
    if (update.name === field.name) {
        hasUpdate = true;
        update.domain = domainFields[i];
    }
    return hasUpdate;
});
```
Wouldn't this accomplish the domain-setting for each matching newField?

jmhauck: @MikeTschudi ahh..you are correct...could do it that way

MikeTschudi: Up to you to change or not; please merge when you want

Issue #59, opened 2019-06-20
Use org's default extent for item extents and feature service initial extent


chris-fox:
When deploying a solution any new items and feature service/view's initial extent should be set to the default extent of the organization/portal.

When implementing this consider we may want to make this an optional parameter to keep the original extent during the deployment, but the default behavior should be to use the org's extent.

chris-fox:
Also just remembered this is a property that Hub templatizes

jmhauck:
@chris-fox If the spatial reference of the ORGs defaultExtent is different than that of the feature service should I be switching it as well?

chris-fox:
I believe the org extent will only ever be WGS 84. We should keep the spatial reference of the service by project the WGS 84 extent into the spatial reference of the service and use that as the service initial extent.

chris-fox:
@jmhauck, I tested this today and found the item extent wasn't being templatized when I use the Create Solutions app. So when I deploy the extent is being set to the same as it was previously in the deployment org. It looks like the latest changes from the app where pushed over. So I am not sure why this isn't working.

chris-fox:
@jmhauck, did some more testing, I found the feature service initial and full extent are being templatized. However, when I deploy to our dev our (https://arcgis4localgov2.maps.arcgis.com/) I am not seeing the correct extent applied, it should be around Naperville, this is what the Pro deployment tool produces:

XMin: -9821384.71421798
YMin: 5117339.12309
XMax: -9797228.38471584
YMax: 5137789.39951188
Spatial Reference: 102100

This is what the online deployment is producing:

XMin: -15000000
YMin: 2700000
XMax: -6200000
YMax: 6500000
Spatial Reference: 102100

Also this issue needs to also handle the extent property for all item types.  This property is always expressed in WGS 84 and is used to set the extent of things like web maps for example. Sometimes this property is null in which case we don't need to handle it.

jmhauck:
Noticed an issue with this when using private orgs. Have a pull in the resolves it. Will move it back once that is merged.

chris-fox:
@jmhauck, I found the item extent is incorrect it looks like you are switching the XMin and XMax. When I manually edit the item in ago-assistant and switch the xmin and xmax it is correct.

Example:

    "extent": [
        [
            -88.0099999998562,
            41.7080000000663
        ],
        [
            -88.22699999983698,
            41.84499999991977
        ]
    ],

[...]

PR #104, opened 2019-09-09
update extent

jmhauck:
#59

### DIFF
--- a/packages/common/src/restHelpers.ts
+++ b/packages/common/src/restHelpers.ts
@@ -127,13 +127,29 @@ export function convertExtent(
                       : undefined;
                   if (projectGeom && projectGeom.rings) {
                     const ring: any = projectGeom.rings[0];
-                    resolve({
-                      xmin: ring[0][0],
-                      ymin: ring[0][1],
-                      xmax: ring[2][0],
-                      ymax: ring[2][1],
-                      spatialReference: outSR
-                    });
+                    resolve(
+                      ring.reduce((accumulator: any, current: any) => {
+                        return {
+                          xmax: Math.max(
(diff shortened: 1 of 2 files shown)


packages/common/src/restHelpers.ts
```diff
@@ -127,13 +127,29 @@ export function convertExtent(
...
                     const ring: any = projectGeom.rings[0];
-                    resolve({
-                      xmin: ring[0][0],
-                      ymin: ring[0][1],
-                      xmax: ring[2][0],
-                      ymax: ring[2][1],
-                      spatialReference: outSR
-                    });
+                    resolve(
+                      ring.reduce((accumulator: any, current: any) => {
```
MikeTschudi: This seems to be more elaborate than we need; the extents to be projected are submitted clockwise starting in the southwestern corner, so the results should be the same. Is there a projection that we have to support for which this doesn't hold true?

jmhauck: I could not answer that. Just so many pulls for one issue. I felt bad that it had come back to my queue yet again and went this way as I thought it was safer and regardless of the order they came back it should produce the expected values.

jmhauck: Part of what threw me off in the first place was they are not returned in the order they are submitted so it made me question what I could rely on

Submitted as:
[[xmin, ymin],[xmin, ymax],[xmax, ymax],[xmax, ymin],[xmin, ymin]]

But returned as:
[[xmax, ymin],[xmin, ymin],[xmin, ymax],[xmax, ymax],[xmax, ymin]]

I'll gladly change back if you think that's best.

PR #105, opened 2019-09-11
Implementation of point-based extents projection

MikeTschudi:
cf. #104, which I suggest that this replaces


MikeTschudi:
arcgis-rest-types continue to be a challenge to access

Issue #86, opened 2019-08-19
Need to templatize Ops Dashboard dataset references


chris-fox:
Operations Dashboard allows you to reference data sources outside the web map as standalone sources. We need to templatize these during the solution creation as well.

A datasource reference will look like below and can be found in the Widgets collections as well as the selectors within the Header or Left Panel.
                                    "dataSource": {
                                        "type": "featureServiceDataSource",
                                        "itemId": "e6f6f748903e4528b004b7c2e4136073",
                                        "layerId": 0,
                                        "table": false
                                    },

Below is the current python code showing how these datasources are found

The Election Results solution has an example where a datasource to a feature service is specified. We should only be templatizing featureServiceDataSources that have an itemId, i.e hosted.

In terms of listing the dependencies. I would like to see the feature service as a dependency of the dashboard only if it is not also a dependency of a web map also referenced in the dashboard. For example in Election Results the feature service is also used in one of the web maps, So I don't think we should list it as a dependency of both the dashboard and the web map.

```
        if 'widgets' in self.data:
            for widget in app_json['widgets']:
                if widget['type'] == 'mapWidget':
                    if 'itemId' in widget:
                        widget['itemId'] = clone_mapping['Item IDs'][widget['itemId']]
                elif "datasets" in widget:
                    for dataset in widget["datasets"]:
                        if "dataSource" in dataset and "itemId" in dataset["dataSource"]:
                            # in some cases the layer ids may have changed when cloning
                            # so we can use a mapping between the old and new to update it
                            for url, cloned_service in clone_mapping['Services'].items():
                                if cloned_service['id'] == clone_mapping['Item IDs'][dataset["dataSource"]["itemId"]]:
                                    # update the layer id
                                    dataset["dataSource"]["layerId"] = cloned_service['layer_id_mapping'][dataset["dataSource"]["layerId"]]
                            # update the item id
                            dataset["dataSource"]["itemId"] = clone_mapping['Item IDs'][dataset["dataSource"]["itemId"]]

        if "headerPanel" in app_json and "selectors" in app_json["headerPanel"]:
            for selector in app_json["headerPanel"]["selectors"]:
                if "datasets" in selector:
                    for dataset in selector["datasets"]:
                        if "dataSource" in dataset and "itemId" in dataset["dataSource"]:
                            # in some cases the layer ids may have changed when cloning
                            # so we can use a mapping between the old and new to update it
                            for url, cloned_service in clone_mapping['Services'].items():
[...]

PR #118, opened 2019-09-30
dashboard field references

jmhauck:
#86


jmhauck:
@MikeTschudi believe that's what it was looking for :)

PR #125, opened 2019-10-16
Created examples for live testing and fixed handling of binary data

MikeTschudi:


packages/common/src/restHelpersGet.ts
```diff
@@ -0,0 +1,598 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * Provides common item fetch functions involving the arcgis-rest-js library.
+ *
+ * @module restHelpers
```
jmhauck: ? restHelpersGet

MikeTschudi: Split out get functions from restHelpers because file was getting too big. Would like to split out another part--such as updates--someday.

jmhauck: I hear you...just wondering if the value  after the module decorator restHelpers should change to restHelpersGet as well. I really don't know exactly what it does but just mentioned it as it was not in sync with the modules name :)

MikeTschudi: You're right--it's an error; will fix

PR #130, opened 2019-10-27
various issues for the alpha project

jmhauck:


packages/common/test/generalHelpers.test.ts
```diff
@@ -72,19 +72,21 @@ describe("Module `generalHelpers`: common utility functions shared across packag
...
       });
     });
-  }
 
-  describe("cleanId", () => {
-    it("should handle empty id", () => {
-      expect(cleanId(null)).toBeNull();
-      expect(cleanId(undefined)).toBeUndefined();
-      expect(cleanId("")).toEqual("");
+    describe("blobToFile", () => {
```
MikeTschudi: blobToFile, blobToText, setProp are not needed individually because they are covered 100% via other tests.

jmhauck: @MikeTschudi ok. I had just started to add placeholder tests for each function in case something more specific came up for that function and we may want to add a really focused test. May be just as easy to add the specific test to other tests that already cover it I suppose.

MikeTschudi: Let's please wait until we need it. Eventually, we hope to have no ignored tests or console.logs in the tests.

jmhauck: ok

PR #134, opened 2019-11-04
Isolate the arcgis-rest-* libraries to the `common` package, continued

MikeTschudi:
Removed remaining `rest_*` calls (cf. #132)

### DIFF
--- a/packages/common/src/featureServiceHelpers.ts
+++ b/packages/common/src/featureServiceHelpers.ts
@@ -20,8 +20,6 @@
  * @module featureServiceHelpers
  */
 
-// ------------------------------------------------------------------------------------------------------------------ //
-
 export {
   queryFeatures as rest_queryFeatures,
   addFeatures as rest_addFeatures
@@ -29,20 +27,18 @@ export {
 
 //#region Imports -------------------------------------------------------------------------------------------------------//
 
-/*import {
-  INumberValuePair,
-  IStringValuePair,
-  IItemTemplate,
-  IDependency,
-  IUpdate,
-  IPostProcessArgs
-} from "@esri/solution-common/src/interfaces";*/
-
 import * as auth from "@esri/arcgis-rest-auth";
-import * as interfaces from "./interfaces";
 import * as generalHelpers from "./generalHelpers";
 import * as templatization from "./templatization";
 import * as restHelpers from "./restHelpers";
+import {
+  IDependency,
+  IItemTemplate,
+  INumberValuePair,
+  IPostProcessArgs,
+  IStringValuePair,
+  IUpdate
+} from "./interfaces";
 
 //#endregion ------------------------------------------------------------------------------------------------------------//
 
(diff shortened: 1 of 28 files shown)


packages/common/src/featureServiceHelpers.ts
```diff
@@ -20,29 +20,25 @@
...
-  INumberValuePair,
-  IStringValuePair,
-  IItemTemplate,
-  IDependency,
-  IUpdate,
-  IPostProcessArgs
-} from "@esri/solution-common/src/interfaces";*/
-
 import * as auth from "@esri/arcgis-rest-auth";
-import * as interfaces from "./interfaces";
```
jmhauck: @MikeTschudi do you have a general rule of thumb or preference when we use one pattern over the other? I had started to switch more to this as I had thought more of your stuff was using this approach.

MikeTschudi: @jmhauck, I have been trying to use `import * as x from y` to avoid name clashes and to simplify going back to the source item--I know where to look. I was having trouble importing and exporting with the interface symbols, so I fell back to this form. With a fresh node_modules build, I see that it is not a problem, so I will PR undoing this change. Thanks for asking.

packages/common/src/restHelpers.ts
```diff
@@ -638,10 +656,21 @@ export function getFeatureServiceProperties(
   });
 }
 
+export function searchGroups(
+  searchString: string,
+  authentication: UserSession
+): Promise<ISearchResult<IGroup>> {
+  const searchOptions: portal.ISearchOptions = {
```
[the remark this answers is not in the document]
MikeTschudi: Similar to previous question: problems with builds.

packages/common/src/featureServiceHelpers.ts
jmhauck: @MikeTschudi ah gotcha

packages/simple-types/src/simple-types.ts
```diff
@@ -149,8 +149,8 @@ export function convertItemToTemplate(
               break;
             case "form":
               itemTemplate.dependencies = itemTemplate.dependencies.concat(
-                relatedItemsResponse.relatedItems.map(
-                  relatedItem => relatedItem.id
+                (relatedItemsResponse as any).relatedItems.map(
```
jmhauck: This line confuses me. If it's any how do we know it has a relatedItems array?

jmhauck: @MikeTschudi since we are now at TypeScript 3.7 can we use the new optional chaining for stuff like this? I have not played with it yet just saw someone mention in the devs slack.

`(relatedItemsResponse as any).relatedItems?.map(`

jmhauck: I see the catch that will ensure it's there on error...does the portal query also garuntee it? Just curious.

MikeTschudi: Thanks for pointing out optional chaining--that'll be useful. In this change, the statement compiles in VS Code, but fails during `npm build`, so I added the cast to `any`.

MikeTschudi: What are you referring to for the catch & portal query, please?

jmhauck: @MikeTschudi I was just talking about the fact that I could see that the catch handling would guarantee that it had a relatedItems array but couldn’t see that it necessarily would with the underlying portal query.

`Promise.all([
    dataPromise,
    resourcePromise,
    relatedPromise.catch(
        () =>
        ({ total: 0, relatedItems: [] } as common.IGetRelatedItemsResponse)
    )
]).then(`

Issue #128, opened 2019-10-21
Support templatizing and deploying item types with a download


chris-fox:
For any item types with a download we need to store the download as a resource of the solution item template and then support deploying that with the new created item as it download. Examples:

ArcGIS Pro Project Package
ArcGIS Pro Add In
Desktop Add In
Desktop Application Template
Geoprocessing Sample
Microsoft Word
PDF

chris-fox:
@MikeTschudi, based on your comments on the Election Outreach project package failing to open in Pro 2.4, I tested it as well and found the resulting Pro Project item had no download. I checked the template and the pro project was not being stored in the item's resource, all I found were pngs for the thumbnails and other item resources. 

I wanted to capture this issue to handle all download item types.

chris-fox:
@MikeTschudi, could you take a look at this issue next? I am not finding the download items types are properly having their downloads stored as resources in the template and as a result it is creating new items with no downloads.

chris-fox:
@MikeTschudi, I created the Election Outreach solution today and found that the Solution Icon was being templatized and the Code Attachments for the WAB apps were also being templatized. I am assuming this is related to the work with this issue. Wondering if we could ignore Code Attachments during solution creation (We are already handling creating them for WAB during deployment). Also could we ignore the Solution Icon items (items specifically with the deploy.thumbnail tag).

MikeTschudi:
@chris-fox, do I understand correctly that we'll not use the deploy.thumbnail experimental approach for Solution Icons? No more special-case items in group.

chris-fox:
@MikeTschudi, I do still want to use the deploy.thumnail tagged item to generate the thumbnail for the Solution Item. The issue is in a recent update I found it is not doing this (The thumbnail is not being set) and it is also templatizing the item in the resulting solution item. I don't want this special thumbnail item included in the Solution Item templates. See highlighted below:

![image](https://user-images.githubusercontent.com/6796763/71118738-aaad6700-21a6-11ea-8343-9c6d2f8586b0.png)

MikeTschudi:
Got it--thanks.  I removed it as part of creation rewrite and will restore it.

chris-fox:
@MikeTschudi, I am not aware but I seem to be encountering a new issue w/ regard to deploying solutions referencing data items. When a solution has a template with a data item in it the deployment seems to stop before the final step of updating the final Solution item json, but seems to create everything correctly.

[...]

PR #144, opened 2019-11-22
Refactored top level of creator package

MikeTschudi:
#128


packages/creator/src/creator.ts
```diff
@@ -21,138 +21,187 @@
...
+    // Get group information
+    Promise.all([
+      common.getGroup(groupId, authentication),
+      common.getGroupContents(groupId, authentication)
+    ]).then(
+      responses => {
+        const [groupInfo, groupItems] = responses;
+
+        // Create a solution from the group's contents, using the group's information as defaults for the solution item
+        const createOptions: common.ICreateSolutionOptions = options ?? {};
```
jmhauck: Nullish coalescing...I like it!

packages/creator/test/creator.test.ts
```diff
@@ -283,4 +253,218 @@ describe("Module `creator`", () => {
...
+
+  describe("createSolutionFromItemIds", () => {
+    it("createSolutionFromItemIds fails to create solution item", done => {
+      const itemIds: string[] = [];
+      const authentication: common.UserSession = MOCK_USER_SESSION;
+      const url =
+        "https://myorg.maps.arcgis.com/sharing/rest/content/users/casey/addItem";
+
+      fetchMock.post(url, { success: false });
+      spyOn(common, "createId").and.callFake(() => "xfakeidx");
```
jmhauck: Interesting...I had to read on this a bit to follow. Thats a cool pattern that will come in handy

packages/creator/src/creator.ts
MikeTschudi: Thanks to you!   :-)

PR #168, opened 2020-01-27
Added Web Scene and give deployed Form its zip file

MikeTschudi:
Also added display of related items to getItemInfo example for testing completeness of item conversion to solution and subsequent deployment.


MikeTschudi:
Deferring web scenes

PR #196, opened 2020-02-08
update groups and circular dependencies approach

jmhauck:


packages/common/src/interfaces.ts
```diff
@@ -258,9 +272,10 @@ export interface IItemTemplateConversions {
     templateDictionary: any,
     destinationAuthentication: UserSession,
     progressTickCallback: () => void
-  ): Promise<string>;
-  postProcessCircularDependencies?(
-    newItemTemplate: IItemTemplate,
+  ): Promise<ICreateItemFromTemplateResponse>;
+  postProcessDependencies?(
+    newItemTemplate: IItemTemplate[],
```
MikeTschudi: Let's please rename the `newItemTemplate` array to `templates` to match function signature, and `authentication` is clearer than the function's `destinationAuthentication`

packages/common/test/restHelpers.test.ts
```diff
@@ -1837,9 +1837,21 @@ describe("Module `restHelpers`: common REST utility functions shared across pack
...
-    xit("shareItem", done => {
-      console.warn("========== TODO ==========");
-      done.fail();
+    it("can handle error on shareItem", done => {
+      const groupId: string = "grp1234567890";
+      const id: string = "itm1234567890";
+      fetchMock
+        .get(
+          "https://myorg.maps.arcgis.com/sharing/rest/community/users/casey?f=json&token=fake-token",
+          MOCK_USER_SESSION.token
```
MikeTschudi: Why `MOCK_USER_SESSION.token` as the response, please? Doesn't match what AGO would return for a user query like this.

packages/common/src/interfaces.ts
jmhauck: good catch..will do

packages/common/test/restHelpers.test.ts
jmhauck: oops...not intended

Issue #175, opened 2020-01-30
Add ability to reuse items already created from the same template (ex. Feature Layer referenced in multiple solutions)


shoe913:
If solutions share a common layer, the layer is deployed for each solution when they aren't deployed with the same solution item.

This same behavior occurs when a solution is deployed twice. This could have repressions depending on what we see the solution update pattern being.

chris-fox:
@MikeTschudi, to support this I am thinking we take this approach:

- Any item that is created from a template during deployment should write a new type keyword to the item using the pattern "source-_itemId_" for example source-f6a4373a92b1471983a063053d43eab2. This should be the template's unique id, and by default will be the same id as the item it was created from

- During template creation we should remove any "source-_itemId_" typeKeywords from the item template. This will ensure that only one source- typeKeyword will be applied to the item created from the template.

- An optional flag should be added to the api to allow the client app to determine whether during deployment items should be search for to see if they have already been created, or if this should be ignored an all new items should be created.

- If the flag is set to search for existing items we should search the user's content for any items that have the matching typeKeywords for any of the templates being deployed. 

- If a matching typeKeyword is found and it is the same item type (Should always be, but is a good safe check) then we should use that item in their org rather than creating a new item. All of the properties of that item should be used in the replacement dictionary so that any items with dependencies on it will use the existing item in their content.

- If multiple items have the same matching typeKeyword we should use the item with the most recent creation date (The item that was deployed most recently.

Using this same typeKeyword pattern will allow us to reuse items that were created with the Pro add-in.

chris-fox:
FYI @shoe913

chris-fox:
@shoe913, Ryan added the logic to the ArcGIS Solutions app. I was testing with the Election Outreach and Election Management solutions and found it was reusing the items from the previous deployment well. I also tested deploying Election Outreach from pro and the Election Management from the app and it also correctly found and used the layer. This is looking good to me, but wanted you to take a look as well.

shoe913:
Looks good to me as well. I think Ryan hit the nail on the head with this. 

I deleted the Solution Item and redeployed. The Solution Item was recreated, but no maps and data were. On another redeploy (with item existing in my content) everything was recreated.

PR #210, opened 2020-02-28
reuse items

jmhauck:


jmhauck:
@MikeTschudi this previously showed as passing :( will update the tests more

packages/common/test/mocks/agolItems.ts
```diff
@@ -708,7 +708,7 @@ export function getAGOLGroup(): any {
     owner: "LocalGovTryItLive",
     description: "Description of an AGOL group",
     snippet: "Snippet of an AGOL group",
-    typeKeywords: ["JavaScript"],
```
MikeTschudi: Why move "JavaScript" from typeKeyword to tag?  For WMAs, e.g., the typeKeyword is set behind the scenes to "JavaScript" whether or not the user sets it also as a tag.

jmhauck: @MikeTschudi that function gets a group. Group doesn't have typeKeywords only tags.

MikeTschudi: Thanks!

jmhauck:
#175

PR #214, opened 2020-03-03
update can deploy webmap with dependencies test

jmhauck:


jmhauck:
This needs some additional work. Not sure why it didn't fail locally but as I started to go through it more I noticed an error that needs to be addressed.

jmhauck:
will issue a new PR

PR #220, opened 2020-03-05
Added tests and miscellaneous changes

MikeTschudi:
Restored coverage testing


packages/simple-types/test/dashboard.test.ts
```diff
@@ -125,6 +126,36 @@ describe("Module `dashboard`: manages the creation and deployment of dashboard i
...
+            type: "serviceDataset"
+          }
+        ]
+      };
+
+      const itemTemplate: common.IItemTemplate = templates.getItemTemplate(
+        "Dashboard"
+      );
+
+      dashboard._getDatasourceDependencies(obj, itemTemplate);
```
jmhauck: @MikeTschudi should these have some sort of expect at the end or ok with the fact that they simply complete?

packages/simple-types/test/dashboard.test.ts
```diff
@@ -194,4 +225,76 @@ describe("Module `dashboard`: manages the creation and deployment of dashboard i
...
   });
+
+  describe("_templatizeByDatasource", () => {
+    it("ignores supplied objs if it is not defined", () => {
+      const updatedList = dashboard._templatizeByDatasource(null, null);
+      expect(updatedList).toBeNull();
+    });
+  });
+
+  describe("_getDatasourceInfo", () => {
```
jmhauck: same here?

packages/common/src/interfaces.ts
```diff
@@ -360,21 +360,6 @@ export interface IItemTemplate {
   estimatedDeploymentCostFactor: number;
 }
 
-export interface ICreateItemFromTemplateResponse {
```
jmhauck: @MikeTschudi should this have been added back somewhere?

jmhauck: oh wait...I see...when I look back at the source...it was a second instance of this

packages/simple-types/test/dashboard.test.ts
MikeTschudi: @jmhauck, these look like merge errors that I made; I'll fix; thanks

Issue #173, opened 2020-01-29
Ignore Living Atlas Feature Services during templatization


shoe913:
If a Living Atlas Layer is in a solution, the solution template cannot be deployed because the layer cannot be cloned. Deployment hangs indefinitely. 

Sample: "Living Atlas Test" 
https://localdeployment.maps.arcgis.com/home/item.html?id=3a052dbe96574a9ca37c3160c3a4269d

In the JSON, the URL parameter and ItemID are being turned into parameters 

"operationalLayers": [
                    {
                        "id": "NDFD_Precipitation_v1_858",
                        "layerType": "ArcGISFeatureLayer",
                        "url": "{{f9e9283b9c9741d09aad633f68758bf6.layer2.url}}",
                        "visibility": false,
                        "opacity": 0.65,
                        "title": "National Weather Service Precipitation Forecast - Cumulative Total",
                        "itemId": "{{f9e9283b9c9741d09aad633f68758bf6.layer2.itemId}}"
                    },"

chris-fox:
@shoe913, Could you try changing that operational layer url property and itemId property back to what they were originally in the web map. Then for that template remove the 'f9e9283b9c9741d09aad633f68758bf6' reference from the depedencies. Finally remove the 'f9e9283b9c9741d09aad633f68758bf6' template from the templates array and save the changes to the json. When you deploy does it create the web map correctly and just reference the living atlas layer.

I ask because one thing I am considering for June is prioritizing Deployment bugs over Template authoring bugs. If for example we have to help the Solution author manually repair their template and it will deploy properly, this is something we could ship with. Its not ideal, but we can continue enhancing the template authoring logic after the first release. Whereas if it won't deploy this is more of a showstopper.

chris-fox:
@jmhauck, could you investigate what it would take to determine if the authoring user has admin access to the item (Is the owner or is an admin in the org the item belongs to). Then what would it involve to either not templatize the item or partially templatize it as an external resource depending on if this would need to before or after the depdencies and variables have been set on the parent item. Would like to discuss more after this investigation to decide on the best approach.

shoe913:
@chris-fox 

Those changes did get the item to deploy correctly. 

I agree, for the first release we can certainly have our folks manually make these changes to the JSON. Since we couldn't do this with the pro addin, I think only new solutions will need this and that will be a small number.

chris-fox:
@jmhauck, There is a property returned on the item called itemControl, it is optional. Per the documentation:

[...]

PR #225, opened 2020-03-09
validate group designations

jmhauck:
#173


packages/common/src/restHelpers.ts
```diff
@@ -24,6 +24,7 @@ import * as generalHelpers from "./generalHelpers";
 import * as interfaces from "./interfaces";
 import * as portal from "@esri/arcgis-rest-portal";
 import * as request from "@esri/arcgis-rest-request";
+import * as restHelpersGet from "./restHelpersGet";
```
MikeTschudi: Not needed?

PR #253, opened 2020-03-19
fix id issue with _updateGroupReferences and add test

jmhauck:


MikeTschudi:
Tests pass in Travis CI, but coveralls step failed in it

PR #263, opened 2020-03-31
Added Codecov

MikeTschudi:
Switched to Codecov to match hub.js and arcgis-rest-js repos

May need override merge to switch out coveralls


MikeTschudi:
Abandoned conversion for now

Issue #195, opened 2020-02-07
Item URL is not being correctly updated often (More often seen in Web AppBuilder)


shoe913:
Still trying to find commonalities in problematic apps. 

When this happens, everything appears to swizzle correctly, but the URL still contains the Item ID variable

ex.
https://andyshoe.maps.arcgis.com/apps/webappviewer/index.html?id={{4ff9a7d94f8046e59a97ef57d9d189d0.itemId}}

Problematic:
94a7166a8b0a4e8b81283f69f67d5527
038c21a34f2d482baac8857223cb1bae

Ok:
3c61f4a0bb334009a1802f594890d289
8f16bf59986b4135b1212a0da56d37eb

shoe913:
When two solutions are deployed without refreshing the page, the second solution's itemID is not swizzled for Web AppBuilder applications. 

Deployment appears correct, however the cannot be used because the itemID has not swizzled and is listed as the variable (ex. {{4ff9a7d94f8046e59a97ef57d9d189d0.itemId}} )

chris-fox:
I am ok with a temporary fix of adding a delay between retry updates for now. 

However, I think this could be improved if we try to combine our update calls. Currently for every WAB app we are making 2 back-to-back update calls, first to update the item url and then to update the item data. I believe the second update is overwriting the first. I think if we could do this is a single update this would help significantly. So far I have only seen this issue with WAB.

chris-fox:
Verified

PR #269, opened 2020-04-01
Further effort to get around AGO update URL problem

MikeTschudi:
#195


MikeTschudi:
@jmhauck, Travis is [passing](https://travis-ci.org/github/Esri/solution.js), but not reporting...

Issue #285, opened 2020-04-10
Concurrent deployments of Solutions with a group with same name fails


chris-fox:
Deploy the Coronavirus Response solution 2x at same time and one will fail creating the group:

{"error":{"code":400,"messageCode":"COM_0044","message":"Unable to create group.","details":["You already have a group named 'Public Place Manager'. Try a different name."]}}

https://localdeployment.maps.arcgis.com/home/item.html?id=caf7f1effad04817942d8b619f28ed43
http://ec2-18-219-57-96.us-east-2.compute.amazonaws.com/corona-virus-response/?sortField=Newest#solutions

I believe this is a similar issue to #243. Assuming we are getting all the Group names at the beginning and then it fails when one of the deployments creates the group with that name. Could we implement the same basic solution as #243, where we get the groups and try the first name available and retry if there is an error with a new name?

chris-fox:
@MikeTschudi, just found this one as well after Ryan implemented concurrent deployments in the app. Could you take a look at this one next?

chris-fox:
Verified

PR #286, opened 2020-04-10
Handle concurrent group creation

MikeTschudi:
#285

### DIFF
--- a/packages/common/src/restHelpers.ts
+++ b/packages/common/src/restHelpers.ts
@@ -392,17 +392,6 @@ export function createFullItem(
   });
 }
 
-export function createGroup(
-  groupItem: any,
-  authentication: interfaces.UserSession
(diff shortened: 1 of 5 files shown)


packages/common/src/restHelpers.ts
```diff
@@ -524,6 +513,63 @@ export function createUniqueFolder(
...
+ * Creates a group using numeric suffix to ensure uniqueness.
+ *
+ * @param title Group title, used as-is if possible and with suffix otherwise
+ * @param templateDictionary Hash of facts: org URL, adlib replacements, user
+ * @param authentication Credentials for creating group
+ * @return Information about created group
+ */
+export function createUniqueGroup(
+  title: string,
+  groupItem: interfaces.IGroupAdd,
```
jmhauck, line 526: @MikeTschudi groupItem is missing from the doc

Issue #282, opened 2020-04-09
Reduce back to back updateDefinition calls for WAB 


chris-fox:
Currently for every WAB app we are making 2 back-to-back update calls, first to update the item url and then to update the item data. I believe the second update is overwriting the first.

I think if we could do this in a single update this could help prevent the occurrences of the url not being updated every after returning succesful. So far I have only seen this issue with WAB.

chris-fox:
Verified

PR #287, opened 2020-04-11
Combine calls to minimize updates

MikeTschudi:
#282


packages/simple-types/src/notebook.ts
```diff
@@ -77,11 +77,21 @@ export function fineTuneCreatedItem(
   templateDictionary: any,
   authentication: common.UserSession
 ): Promise<void> {
-  const data: any = common.replaceInTemplate(
-    originalTemplate.data,
-    templateDictionary
-  );
-  return _updateItemData(newlyCreatedItem.itemId, data, authentication);
```
jmhauck, line 84: @MikeTschudi was this function removed as well?

Issue #234, opened 2020-03-11
Security Scans for Creation, Deployment, and Solutions App Use


shoe913:
Add HTML sanitizers to ensure clean solution templates. This will be done on template creation and solution deployment. 

Template: Description, Snippet
Info Products: Description, Snippet, Pop-up codes

chris-fox:
@shoe913, Makes sense for description and snippet. I worry about changing the pop-ups thought. The web map viewer should do this automatically when you save. I don't want to break the pop-ups as part of this process.

shoe913:
@chris-fox 

Our concern with the pop-ups is in regard to users injecting their own html (similar to our standardizing processes) into their Solution's JSON. If this happens anything can be put in and won't be sanitized until the map is opened and saved. 

This may be a fringe scenario, but the fact that we are already doing a similar workflow proves it could be a potential hole. 

I'm not aware of workflows in place which prevent this at REST to the webmap, however those maps won't potentially end up in another organization.

chris-fox:
@shoe913, how are we going to know what is valid and what is invalid html? AGO has specific rules that I am not sure we have ensure we are following their rules and not over doing it or under doing it. Also this is something users can easily do today to their own webmaps so I am not sure why we feel we need to solve this problem.

dbouwman:
Fwiw - I've done a little poking around w/ the [jsxss](https://jsxss.com/en/index.html) lib that backs the [arcgis-html-sanitizer](https://github.com/Esri/arcgis-html-sanitizer) module, and it can be fed a object, and it applies the sanitization rules to all string fields down the whole graph. 

We are looking at switching to this in Hub because it is a more wholistic solution we can apply on every save and load vs handling sanitization at edit time on specific card properties. 

We have more testing to do to ensure we can extend it for aria bootstrap props not in the base lib's rules, and that it does not mess up other stuff... but I think it may be a clean option for Solutions - just run the entire solution /data thru it before deploying

MikeTschudi:
@shoe913, PR #295 has initial solution scanning

MikeTschudi:
#321 is caused by blindly scanning the data section of items; as @dbouwman wrote, "TODO: we should delegate data sanitization to the type-specific modules"

With PR #325, restricting scanning to the item base info; issue #324 takes over data section sanitization.

shoe913:
Verified

PR #295, opened 2020-04-15
Scan various parts of a solution and its items for undesirable content

MikeTschudi:
For issue #234.

PR #279 incorporated [arcgis-html-sanitizer](https://github.com/esri/arcgis-html-sanitizer) via a thin interface.

The solution-deployment-app can use PR #279 to [fetch the solutions in a solution group and screen the solutions' base sections](https://github.com/Esri/solution.js/blob/develop/packages/common/examples/sanitizeGroupItems/src/main.ts), which are what is used to generate the item cards in the app's UI.

This PR uses PR #279 to
* scan the base section properties (the properties that every AGO item has) before creating a Solution Template
* scan the base section of every item that is going into a solution
* scan the data section of every JSON item that is going into a solution: item types Dashboard, Feature Service, Notebook, Project Package, Web Map, Web Mapping Application, and Workforce Project
* scan the base and data sections of a Solution Template before it is used to create a Deployed Solution and to deploy the items in that Solution Template
* scan JSON files stored as resources in the Solution Template because some items store their JSON that way rather than embedded in the Solution Template (e.g., Form info files)

Changes to JSON caused by scans are echoed into the console for now to help us to check what the scanner removes.

---

Two additional changes in this PR:

* Check that a solution id is supplied to deploySolution
* Simplified repo publishing step

---

@dbouwman, thank you for the jsxss investigations--the reassurance is very helpful.


packages/deployer/src/deployer.ts
```diff
@@ -69,10 +73,25 @@ export function deploySolution(
...
+            deployOptions.description,
+            sanitizer
+          );
+
           _deploySolutionFromTemplate(
             templateSolutionId,
-            itemBase,
-            itemData,
+            common.sanitizeJSONAndReportChanges(itemBase, sanitizer),
+            common.sanitizeJSONAndReportChanges(itemData, sanitizer),
```
dbouwman, line 94: @jmhauck sorry I did not see this PR earlier - I'm just reading through this and it looks like the deployer is handling the sanitize vs delegating to the type-specific functions... is that something we could change? I ask because both Hub and StoryMaps override the base sanitizer rules to allow more properties - mainly stuff for accessibility - and running those solution templates through the base sanitizer will break them. I'll also reach out via email as well

Issue #305, opened 2020-05-01
Pass either templateId or templateItem into deploySolution


dbouwman:
Allow consumers of the deployer to pass in the Solution as an object vs assuming it's an item to be fetched.

This is needed for Hub as we have "embedded" solution templates that ship as json files in the Hub application. We also leverage the Solution system as a means to "clone", by simply piping the output from the "create template" process, into the "deploy" process without writing out an item.

dbouwman:
Implemented in #306 - Closing

PR #306, opened 2020-05-01
[DNM] Allow Solution Model or Id to be passed into deploySolution

dbouwman:
Addresses #305 but even if ya'll happy with it, pls do not merge it yet :)

**Note**: this spiraled a bit, and I'm not at all fixed on the naming anything changed in this, or location of the additional util fn's I created... so ya'll likely have better ideas where some of this stuff should live or what it should be named.

### Description

Changes the `solutionItemId: string` param of `deploySolution(...)` to `maybeModel: string| IModel`, and handles both cases.

Adds `_getSolutionTemplateItem(..)` that takes the `maybeModel` and either fetches the item + data if it's a string, or returns the passed in model, if it's an `IModel`.

This PR also refactors `deploySolution(...)` to use a flat promise chain, and hoists out much the between-steps logic into helpers in `deployerUtils`. 

### Adding Sinon
Since I was working in the main entry-point for deploying a solution, I wanted to be able to test the "orchestration" that happens inside that function, vs the whole chain of things that happen when it calls out to other fns. To that end, I brought in Sinon to give us a mock/stub system.
In order to make this work, I had to move `_deploySolutionFromTemplate` into it's own .ts file.
In `deployer.ts` I imported just the function a la

```
import { _deploySolutionFromTemplate } from "./deploySolutionFromTemplate";
```

Then in the `deployer.test.ts` I imported the same thing via 

```
import * as deploySolutionFromTemplate from "../src/deploySolutionFromTemplate";
```

Which then allows...
```
// in deployer.test.ts
...
     deployFnStub = sinon
          .stub(deploySolutionFromTemplate, "_deploySolutionFromTemplate")
          .resolves("3ef");
...
     expect(deployFnStub.calledOnce).toBe(
              true,
              "_deploySolutionFromTemplate should be called once"
            );
```

I did the same thing with the `_getSolutionTemplateItem`, and `getItemMetadataAsFile` fns allowing us verify the "orchestration" of `deploySolution(...)` without also mocking all the calls down the entire stack. 

That said, all the old tests are still present and running as well (heroic fetch-mocking btw!)

There is a lot of things moved in this, but we're still at 100% coverage. 

If you want, I can walk ya'll thru this on Monday.

### DIFF
--- a/packages/deployer/src/deployer.ts
+++ b/packages/deployer/src/deployer.ts
@@ -21,458 +21,93 @@
  */
 
 import * as common from "@esri/solution-common";
-import * as deployItems from "./deploySolutionItems";
 
-//#region Entry point ----------------------------------------------------------------------------------------------- //
+import { _deploySolutionFromTemplate } from "./deploySolutionFromTemplate";
+import {
+  _getSolutionTemplateItem,
+  isSolutionTemplateItem,
+  _updateDeployOptions
+} from "./deployerUtils";
+import { IModel } from "@esri/hub-common";
 
+/**
+ * Deploy a Solution
+ * Pass in either the item id or an IModel (`{item:{}, model:{}}`)
+ * of a Solution Template, and this will generate the Solution
+ * @param maybeModel Item Id or IModel
+ * @param authentication UserSession
+ * @param options object
+ */
 export function deploySolution(
-  templateSolutionId: string,
+  maybeModel: string | IModel,
   authentication: common.UserSession,
   options?: common.IDeploySolutionOptions
 ): Promise<string> {
-  return new Promise((resolve, reject) => {
-    if (!templateSolutionId) {
-      reject(common.fail("The Solution Template id is missing"));
-    }
+  // if we are not passed the maybeModel, reject
+  if (!maybeModel) {
+    return Promise.reject(common.fail("The Solution Template id is missing"));
+  }
+  let deployOptions: common.IDeploySolutionOptions = options || {};
+  /* istanbul ignore else */
+  if (deployOptions.progressCallback) {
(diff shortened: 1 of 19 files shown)


dbouwman:
Unclear why/how this is failing in CI ... it does the `npm install` and then it runs `npm run test:ci` and in that, it reports this error...

```
Error: Cannot find module '@esri/arcgis-rest-portal'
```

packages/deployer/src/deployer.ts
```diff
@@ -21,458 +21,93 @@
...
           deployOptions.progressCallback(1);
         }
-        reject(error);
+        return Promise.reject(error);
       }
-    );
-  });
+    )
+    .catch(ex => {
+      throw ex;
```
MikeTschudi, line 111: Please change to Promise.reject--the several current clients of the library only have to support Promise returns

packages/deployer/test/deployerUtils.test.ts
```diff
@@ -0,0 +1,125 @@
...
+import * as common from "@esri/solution-common";
+import * as testUtils from "@esri/solution-common/test/mocks/utils";
+
+let MOCK_USER_SESSION: common.UserSession;
+
+beforeEach(() => {
+  MOCK_USER_SESSION = testUtils.createRuntimeMockUserSession();
+});
+
+describe("Module: `_deployerUtils", () => {
```
MikeTschudi, line 32: single backtick?

dbouwman: Whoops! Thanks!

packages/deployer/src/deployer.ts
dbouwman: That will return a rejected promise :)

PR #315, opened 2020-05-05
item id getting lopped off

jmhauck:


packages/deployer/src/deployer.ts
```diff
@@ -82,7 +82,7 @@ export function deploySolution(
       common.deleteItemProps(item);
 
       return _deploySolutionFromTemplate(
-        item.id,
+        deployOptions.jobId,
```
MikeTschudi, line 85: The jobId is an app-specified id, not a solution id; it could be a random number or any old string. If not supplied by the app, we set it to an id. It's carried along and reported back to the app along with a progress percent.

jmhauck: @MikeTschudi ok, thanks...I had wondered. I saw that it was set to the solution item id in _updateDeployOptions if it didn't have anything.

Issue #310, opened 2020-05-04
Add optional 'progressEvent' property to ISolutionProgressCallback


cosbyr:
Goal: Report back to the Solution Deployment App when a Item that is part of a Solution Item's data.templates list is finished

Proposed new `ISolutionProgressEvent` Structure:
```ts
type interface ISolutionProgressEvent
{
      event: string,
      data?: any
}
```

Proposed Updated `ISolutionProgressCallback` Structure:
```ts
type interface ISolutionProgressCallback
{
      percentDone: number,
      jobId?: string,
      progressEvent?: ISolutionProgressEvent
}
```

progressEvent in the case of reporting back when a Item that is part of a Solution Item's data.templates list is finished:
```ts
{
      'event': 'solutionDataTemplateItemFinished',
      'data': 'item id of source data template id from which the new item was created'
}
```

I think this ISolutionProgressEvent interface is something we can possibly use for other types of events moving forward

MikeTschudi:
@cosbyr, we have status strings that are used in the console progress messages:

```typescript
export const SItemProgressStatus = [
  "1 Started",
  "2 Created",
  "3 Cancelled",
  "3 Finished",
  "3 Ignored",
  "3 Failed"
];
```

Would these be OK for the `event` property?  (The numeric prefixes are for sorting the console log output and can be removed.)

MikeTschudi:
At this point, however, it appears that you only want the "2 Created" message with the created item's id--correct?

cosbyr:
@MikeTschudi I didn't do a good job of making this clear, but I actually want the source template item's ID that the new item was created from once the new item is finished getting created. i'm okay with a different event name.

cosbyr:
@MikeTschudi it might not be "Created" that we want, it might be "Finished", the item needs to not just be created but needs to have all of its data in place too before it reports back on the callback i am thinking of.

shoe913:
Verified

PR #317, opened 2020-05-06
Added progress event

MikeTschudi:
#310 

@cosbyr, will this work for you?

### DIFF
--- a/packages/deployer/src/deploySolutionItems.ts
+++ b/packages/deployer/src/deploySolutionItems.ts
@@ -234,13 +234,21 @@ export function deploySolutionItems(
       itemId: string,
       status: common.EItemProgressStatus,
       costUsed: number,
-      createdItemId: string // supplied when status is EItemProgressStatus.Created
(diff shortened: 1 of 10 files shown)


packages/deployer/src/deploySolutionItems.ts
```diff
@@ -234,13 +234,21 @@ export function deploySolutionItems(
...
     ) => {
       // ---------------------------------------------------------------------------------------------------------------
       percentDone += progressPercentStep * costUsed;
       /* istanbul ignore else */
       if (options.progressCallback) {
-        options.progressCallback(percentDone, options.jobId);
+        if (status === common.EItemProgressStatus.Finished) {
+          const event = {
+            event: common.SItemProgressStatus[status],
+            data: createdItemId
```
cosbyr, line 246: On this line I believe I want the "itemId" not the "createdItemId". I want the ID of the template Item that the createdItem was created from.

MikeTschudi: sorry--misread spec. thanks for catch

PR #326, opened 2020-05-11
F/268 solution schema versioning (Take 2)

dbouwman:
### Take two...
in which we have resolved the upstream issues in hub.js so that:
a) Blobs are only handled in browsers, and this eliminates the `stream` import
b) the node typescript `util` library is not used, thus eliminating the `util` import
c) we have resolved some of the import paths, resolving the circular dependency 

Locally, this also changes the UMD build to treat the `@esri/hub-*` packages as external imports, so UMD packages using solution-js will also need to bring in the `hub-common` umd package. I'm not sure how that works in terms of changed to other apps, but in order to support sites and pages, we will be bringing in more of the `@esri/hub-*` packages. 

This builds clean, and passes all tests w. 100% coverage for me on 12.16.1, and 10.16.1

### Original PR Description
--- 
This is just setting up the schema versioning system.

Adds `item.properties.schemaVersion` and runs migrations if needed.

Currently it just stamps schemaVersion as 3 unless the app encounters a solution w/ Hub specific typeKeywords - which it should not b/c the hub specific items lack the `Template` keyword.

It is one connected into the main workflows in `deployer` - but ideally we'd have a fn `getSolutionItem(...): Promise<ISolutionItem>`, put the migration in there and have everything use that fn.

### DIFF
--- a/packages/common/test/libs.test.ts
+++ b/packages/common/test/libs.test.ts
@@ -217,11 +217,11 @@ describe("Module `arcgis-html-sanitizer`: ", () => {
     });
 
     it("tests XSS cases", () => {
-      console.log(
-        "Running " +
-          xssFilterEvasionTestCases.testCases.length +
-          " XSS test cases"
-      );
+      // console.log(
+      //   "Running " +
+      //     xssFilterEvasionTestCases.testCases.length +
+      //     " XSS test cases"
+      // );
       const sanitizer = new libs.Sanitizer();
 
       xssFilterEvasionTestCases.testCases.forEach(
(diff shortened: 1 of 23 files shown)


packages/common/test/libs.test.ts
```diff
@@ -217,11 +217,11 @@ describe("Module `arcgis-html-sanitizer`: ", () => {
     });
 
     it("tests XSS cases", () => {
-      console.log(
-        "Running " +
-          xssFilterEvasionTestCases.testCases.length +
-          " XSS test cases"
-      );
+      // console.log(
```
MikeTschudi, line 220: Just to know, this is here because I swap out the basic tests with a more complete set of tests, and this lets me verify that I'm testing with 143 cases instead of just the basic 6. But it does insert an untidiness in the output, so I'll remove it.

PR #332, opened 2020-05-14
Hub Type Stubs and Deployer Demo App

dbouwman:
## Update 2
---
- pull `tslib` from node_modues
- update ui so you can paste in a Solution Id or pick from a list

---
## Update
---
- Adds a README with instructions for the demo/deploySolution app
- stubbed in HubSiteProcessor and HubPageProcessor
- minor changes to some deployer functions
- in tracing through the deployer code I see many opportunities to flatten out the promises, and refactor out blocks into separate functions. I'll be submitting smaller PRs around that shortly

---

This stubs in some stuff for hub types, but it also moved the deployer example up to /demos/deploySolution, sets it up w/ package.json and sets up lerna to install all the deps, and then load the umd's from node_modules/**...

This saves copying files into the /lib folder(s) every time you want to test it

Instructions: (will add to a readme once we nail this down)
- pull, `npm install` in the root - this should install the deps for /demos/deploySolution
- in demos/deploySolution, run `npm run build` - once we verify this, we'll move it up to the root
- after the build, in the same location run `npm run start` - this will fire up a web server, and then you can open http://127.0.0.1:8080/ and view the app

Also:
- loads require from it's cdn
- loads the bootstrap css to make the example app look pretty 🌷 

image

image'

Using a solution that works...
image

I'd like @tomwayson to pull this and check my work here, but I *think* this a) works, and b) is reasonably sane

@MikeTschudi - I think this still accomplishes a) testing the libs and b) verifying that the UMD output is solid... but as I've said, I've never done `require` based dev't sooo... pls double check things :)

Once we get this sorted out, we can move the other examples into `/demos` and wire them up the same way, then add more scripts to automate start up etc.


dbouwman:
soooo bizzaro - this was working last night, and now I get errors resolving the `solution-*` packages... looking into it

tomwayson:
This LGTM. I pulled it, tried it, it worked as advertised. A few recommendations:

- Add a README.md to the demo w/ the above intructions
- remove tslib.js, it's not used

Also, I assume the purpose of this demo is to verify that the UMD output of the solution packages works in an AMD loader, right? If you're just trying to verify that the functions work, I'd recommend using a bundler like parcel, rollup or webpack to output a script you can load in index.hml w/o `require()`.

dbouwman:
ok - at least locally - the deal was to set the `module` format to `umd` as that uses "node style" package resolution, which finds the files in... 🥁 `node_modules`

More info on typescript module resolutions https://www.typescriptlang.org/docs/handbook/module-resolution.html

tomwayson:
that's odd b/c the AMD build worked for me locally this AM.

demos/deploySolution/tslib.js
```diff
@@ -0,0 +1,243 @@
+/*! *****************************************************************************
```
dbouwman, line 1: I'm not clear if we need this or not?

MikeTschudi: I haven't been able to definitively say that we don't need it, so I've left it in.

tomwayson: I'm almost certain it is not needed. I don't see anything that loads it. Easy way to find out is remove it and run the demo. If it don't break, it ain't needed.

MikeTschudi: In the PR commit, if I omit tslib.js, the app will not load.

dbouwman: ok - let's leave it for now

tomwayson: ok, I guess if the path is not defined, require.js just defaults to `./`. I'd say it would be better to define a path to tslib in node_modules the way you have w/ other packages.

packages/deployer/src/deploySolutionItems.ts
```diff
@@ -584,16 +587,29 @@ export function _createItemFromTemplateWhenReady(
...
-        awaitDependencies.push(templateDictionary[dependencyId].def);
-      });
+      // maybePush will only add the entry if it exists, allowing
+      // templates to have entries in the depenedency array that are
+      // invalid or don't refer to a template
+      const awaitDependencies = template.dependencies.reduce((acc, id) => {
+        return maybePush(templateDictionary[id].def, acc);
+      }, []);
+
+      // const awaitDependencies = [] as Array<Promise<common.ICreateItemFromTemplateResponse>>;
```
MikeTschudi, line 604: still need former method?

packages/deployer/src/deploySolutionItems.ts
```diff
@@ -694,6 +710,33 @@ export function _createItemFromTemplateWhenReady(
   return itemDef;
 }
 
+// export function _createItemFromTemplateWhenReadyNew(
```
MikeTschudi, line 713: Cut?

packages/deployer/src/deploySolutionItems.ts
dbouwman: ah - no...

packages/deployer/src/deploySolutionItems.ts
dbouwman: yeah - sorry - I should have nixed this stuff out

Issue #313, opened 2020-05-05
Ignore default extents when creating items if extents are invalid


shoe913:
When an organization has an invalid default map extent, the solution.js will not create the solution items correct and fail. 

When the code encounters this, we should handle for for the error and default the extent of the full base map.

![image](https://user-images.githubusercontent.com/11777080/81100940-30198580-8ec2-11ea-9a95-3bd41a09e666.png)

chris-fox:
Verified.

PR #336, opened 2020-05-18
extent failover

jmhauck:
Some autoformatting happened there when I comitted.

@chris-fox @MikeTschudi I used this as the default:

    extent.xmin = -179;
    extent.xmax = 179;
    extent.ymax = 89;
    extent.ymin = -89;
    extent.spatialReference = { wkid: 4326 }

If I used the 180/90 project would return NaN values. If you know of a more suitable default please let me know.

### DIFF
--- a/packages/common/src/restHelpers.ts
+++ b/packages/common/src/restHelpers.ts
@@ -337,7 +337,7 @@ export function convertExtent(
   outSR: ISpatialReference,
   geometryServiceUrl: string,
   authentication: UserSession
-): Promise<IExtent> {
+): Promise<any> {
   const _requestOptions: any = Object.assign({}, authentication);
   return new Promise<any>((resolve, reject) => {
     // tslint:disable-next-line:no-unnecessary-type-assertion
--- a/packages/common/src/restHelpers.ts
+++ b/packages/common/src/restHelpers.ts
@@ -20,11 +20,81 @@
  * @module restHelpers
  */
 
-import { appendQueryParam, blobToJson, blobToText, checkUrlPathTermination, deleteProp, deleteProps, fail, getProp, getUniqueTitle, setCreateProp } from "./generalHelpers";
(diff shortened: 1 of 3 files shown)


jmhauck:
#313

packages/common/src/restHelpers.ts
```diff
@@ -20,11 +20,81 @@
...
-import { IAddFolderResponse, IAddGroupResponse, IAdditionalSearchOptions, ICreateItemResponse, ICreateServiceResult, IDependency, IExtent, IFeatureServiceProperties, IFolderStatusResponse, IGroup, IGroupAdd, IItem, IItemTemplate, IItemUpdate, IPostProcessArgs, IRelatedItems, ISpatialReference, IStatusResponse, ItemRelationshipType, IUpdate, IUpdateItemResponse, UserSession } from "./interfaces";
-import { addItemData as portalAddItemData, addItemRelationship, addItemResource, createFolder, createGroup, createItemInFolder, getItem, IAddItemDataOptions, ICreateItemOptions, IFolderIdOptions, IGroupSharingOptions, IItemResourceOptions, IManageItemRelationshipOptions, ISearchGroupContentOptions, ISearchOptions, ISearchResult, ISetAccessOptions, ISharingResponse, IUpdateItemOptions, IUserGroupOptions, IUserItemOptions, removeFolder as portalRemoveFolder, removeGroup as portalRemoveGroup, removeItem as portalRemoveItem, searchGroupContent, searchGroups as portalSearchGroups, searchItems as portalSearchItems, SearchQueryBuilder, setItemAccess, shareItemWithGroup, updateItem as portalUpdateItem } from "@esri/arcgis-rest-portal";
+import {
```
MikeTschudi, line 23: thanks!

packages/common/src/restHelpers.ts
```diff
@@ -180,7 +337,7 @@ export function convertExtent(
   outSR: ISpatialReference,
   geometryServiceUrl: string,
   authentication: UserSession
-): Promise<IExtent> {
```
MikeTschudi, line 183: What was the problem with the type, please?

chris-fox:
looks good.

packages/common/src/restHelpers.ts
jmhauck: @MikeTschudi If the project call does not return an array with two things it can resolve undefiend. I just thought any was more fitting. However, I'm not sure that would ever be hit. It seems like calls that do not fail but also do not produce valid output just return "NaN" coordinate values for our points.

PR #340, opened 2020-05-21
Support creating/deploying Survey123 published surveys

rweber-esri:
Survey123 `Feature Service` items have a `typeKeyword` that references the related `Form` item ID. However, the `Feature Service` items are dependencies of the `Form` item, so we need to post process the `Feature Service` items after the `Form` item is created. Additionally, only item data is currently post-processed, this PR introduces the ability to post-process item properties as well.

### DIFF
--- a/packages/deployer/src/deploySolutionItems.ts
+++ b/packages/deployer/src/deploySolutionItems.ts
@@ -28,6 +28,7 @@ import * as simpleTypes from "@esri/solution-simple-types";
 import * as storyMap from "@esri/solution-storymap";
 import { HubSiteProcessor, HubPageProcessor } from "@esri/solution-hub-types";
 import { getProp, maybePush } from "@esri/hub-common";
+import * as deployUtils from "./deployerUtils";
 const UNSUPPORTED: common.moduleHandler = null;
 /**
  * Mapping from item type to module with type-specific template-handling code.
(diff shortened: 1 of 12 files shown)


packages/deployer/src/deploySolutionItems.ts
```diff
@@ -790,39 +791,42 @@ export function postProcessDependencies(
...
 
-    const dataRequests: Array<Promise<any>> = [];
+    const modelRequests: Array<Promise<any>> = [];
     const requestedItemInfos: any = clonedSolutionsResponse.filter(
       solutionInfo => {
         if (solutionInfo.postProcess) {
-          dataRequests.push(
-            common.getItemDataAsJson(solutionInfo.id, authentication)
+          modelRequests.push(
+            deployUtils._getSolutionTemplateItem(
```
dbouwman, line 799: I think this will run into issues on two fronts:
a) if the item type does not return a json response, it will throw, and
b) `_getSolutionTemplateItem` applies a schema migration, which is intended for `Solution` items

I'm more inclined to see if we can delegate all of this off to the `itemHandler` - maybe in a generic `postProcess` function. Then it can fetch what it needs, based on the item type and the actions it needs to take

dbouwman: working my way through a refactor of this, I see now that a) above is irrelevant as it's already getting the /data... but we should use a more generic fn than `getSolutionTemplateItem`

rweber-esri:
Closing this PR for now, going to create a new branch from @dbouwman [branch](https://github.com/Esri/solution.js/tree/f/streamline-post-process) that reworked some things in post processing and resubmit

PR #343, opened 2020-05-22
Post Processing delegated to Item Handlers

dbouwman:
This moved a *lot* of stuff around, but the bottom line is that what used to be in the `postProcessDependencies` function is now entirely delegated out a new `postProcess` function exposed on the `itemHandler`s.

This allows each handler to deal with all the post processing, fetching the data or whatever else, doing interpolations, group sharing etc etc. 

This also allows for finer grained tests - some placeholder test suites have been added.

Summary of changes
- `_deploySolutionFromTemplate` now calls `itemHandler.postProcess(...)`
- `simple-types` package exports `simpleTypes` and `notebookProcessor`
   - both have `postProcess(...)`
- `HubSiteProcessor` is connected in, but does not yet implement it's specific logic

Things to resolve:
- processor naming. The simple-types package now exports `notebookProcessor` and `simpleType` - I don't have strong feelings around the specifics, but it would be good to be consistent.

- Notebook isolation: currently the notebook and simple-types processors delegate to two main helper functions `convertItemToTemplate` and `createItemFromTemplate`, and those functions use the `notebook.ts` module, creating a circular ref. To break this, we will need someone who knows the notebook and simple type logic to cleanly separate the two. I can take a crack at it, but only after Sites are working.

Additionally, I removed the `npm install` step from `travis.yml` as travis will run `npm ci` which does the same thing... thus, this cuts the CI run time by ~2 minutes.


packages/simple-types/src/helpers/convert-item-to-template.ts
```diff
@@ -0,0 +1,157 @@
+import * as common from "@esri/solution-common";
+import * as dashboard from "../dashboard";
+import * as form from "../form";
+import * as notebook from "../notebook";
```
dbouwman, line 4: We need to extract notebook from this...

packages/simple-types/src/helpers/create-item-from-template.ts
```diff
@@ -0,0 +1,256 @@
+import * as common from "@esri/solution-common";
+import * as notebook from "../notebook";
```
dbouwman, line 2: We need to extract `notebook` from this as well

packages/simple-types/src/helpers/notebook-helpers.ts
```diff
@@ -0,0 +1,14 @@
+import { generateEmptyCreationResponse } from "./generate-empty-creation-response";
+import { convertItemToTemplate } from "./convert-item-to-template";
+import { createItemFromTemplate } from "./create-item-from-template";
+import { shareTemplatesToGroups } from "./share-templates-to-groups";
+import { updateNotebookData } from "./update-notebook-data";
+
+// Export the Notebook Helper module
+export {
+  createItemFromTemplate,
+  convertItemToTemplate,
```
dbouwman, line 10: re-exporting these two functions was needed to get `spyOn` working smoothly, but I think we can resolve this my renaming the functions so they don't exactly match the fn names on the processors. Additionally, once we hoist out the notebook stuff, this becomes less of an issue, as these fn's will really be just for the "simple types"

packages/simple-types/src/helpers/share-templates-to-groups.ts
```diff
@@ -0,0 +1,36 @@
...
+} from "@esri/solution-common";
+import { maybePush } from "@esri/hub-common";
+
+/**
+ * Given the created templates
+ * @param templates
+ * @param templateDictionary
+ * @param authentication
+ */
+export function shareTemplatesToGroups(
```
rweber-esri, line 15: Curious why we're calling `shareTemplatesToGroups` _only_ from `Notebook` and `SimpleTypes` `postProcess` methods? Is the thinking that different item types may be shared with different groups (or not at all) based on certain criteria?  Or if this should be common to all templates regardless of type, maybe this would make more sense to live in `solution-common` or `solution-deployer` and invoked from `packages/deployer/src/helpers/post-process.ts` instead of individual handler `postProcess` methods that may or may not be implemented?

dbouwman: Good question @rweber-esri ! In general - what I've found from our work building a solution system in Hub, is that delegating as much as you can to type-specific processors is better. Almost everything that I was *sure* would be applied to all item types, I've ended up having to hoist back out because one corner case or another proved me wrong.

For this case specifically, Sites have very particular logic around what groups they are shared to, so I pushed this logic into a shared module, and any/all types that need it, can just import and call that fn. I have no strong feelings for where that specific function lives though.

packages/deployer/test/deploySolutionItems.test.ts
```diff
@@ -1216,359 +1216,359 @@ describe("Module `deploySolutionItems`", () => {
...
-            _templates,
-            clonedSolutionsResponse,
-            MOCK_USER_SESSION,
-            templateDictionary
-          )
-          .then(done.fail, done);
-      });
-    }
-  });
+  // describe("postProcessDependencies", () => {
```
MikeTschudi, line 1219: No longer needed?

packages/simple-types/test/notebook.test.ts
```diff
@@ -1,14 +1,14 @@
 /** @license
  * Copyright 2020 Esri
  *
- * Licensed under the Apache License, Version 2.0 (the "License");
+ * Licensed under the Apache License, Version 2.0 (the 'License');
  * you may not use this file except in compliance with the License.
```
MikeTschudi, line 5: Just to know, the original statement is the format that was specified by Ben Parker and Keith Ryden back in 2012; it wasn't a style choice on our part.

packages/deployer/test/deploySolutionItems.test.ts
dbouwman: tbh - I was not clear around this, as that's a massive set of tests, which I've seemingly replaced (at least in terms of coverage) with way fewer... I wanted to keep this around in case I run into some issue where that code would be helpful... I'll remove it w/ #347 PR

packages/simple-types/test/notebook.test.ts
dbouwman: Thanks for the info - it's always interesting on this stuff - hub.js does not have any licensing info in the individual files, and rest-js just has this header

```
/* Copyright (c) 2018 Environmental Systems Research Institute, Inc.
 * Apache-2.0 */
```

The UMD builds for all of them inject a different license header, the ESM and CJS outputs keep the file headers if they exist... maybe we need to connect with Keith and Ben re: what should be in Hub.js :)

Issue #346, opened 2020-05-26
Quick Capture Project issue with adding resources


shoe913:
When deploying a quick capture project, it will fail with the following error. 

![image](https://user-images.githubusercontent.com/11777080/82946787-d6056080-9f53-11ea-8eda-5d5e1b24c8d3.png)

The resource causing the failure is the qc.project.json. This does exist and is accessible (verified by python api). 

It appears the UpdateResources request doesn't have the correct path for the json file for the package. In the post no sub folder for storing resources is listed

![image](https://user-images.githubusercontent.com/11777080/82946734-bff7a000-9f53-11ea-9c16-330153ab4914.png)

shoe913:
Can reproduce with "Andy Quick Capture"

chris-fox:
Assigned to @jmhauck because he originally did the work

shoe913:
Verified

PR #349, opened 2020-05-28
QuickCapture resources

jmhauck:


packages/simple-types/src/simple-types.ts
```diff
@@ -123,18 +125,24 @@ export function postProcess(
...
-        itemId,
-        { id: itemId },
-        updatedData,
-        authentication
-      ) as Promise<any>;
-    } else {
-      return Promise.resolve({ success: true });
-    }
-  });
+  if (type === "QuickCapture Project") {
```
dbouwman, line 128: I'd consider creating and exporting another "Processor" for quick capture - basically the same way `notebook` is handled. That allows the `simpleTypes` stuff to continue to be super generic, but when we run into types that need special processing, we can isolate that off, which then makes testing simpler as well.

dbouwman: Your call - we're crunched for time, but from my experience having gone through 3 iterations of this in the Hub app, delegating to type-specific handlers has always made things simpler to work with, test, extend and reason about :)

jmhauck:
#346

PR #354, opened 2020-05-29
Hub Site/Page Resources

dbouwman:
**Note:** This moved & refactored some low-level blob functions:
- `add-resource-from-blob.ts`
- `copy-resource.ts`
- `get-blob.ts`

While I'm pretty sure everything is solid (all previous tests pass) everyone working in this project should be watching out for oddities related to fetching blobs or copying resources as blobs.

--- 

Hub Related Changes:
- Handle Resources in pre-existing Hub-created Solution Templates
- Site item get's id's of all created items in it's `item.properties.children` array
- Site and Page processors check state via progress callback 
   - they can early exit
   - they can remove created items if informed at the end

Also addresses:
- more refactoring of fn's out of massive modules into separate files, which enable consistent use of `spyOn` vs having to let code run through all the functions during tests (aka this supports test isolation)
- when moving fns to separate files, also flattened promise chains
- when a solution is loaded we check that each template has `estimatedDeploymentCostFactor` value, if not we set it to 1
- located source of unmocked calls during tests (deployer::deploySolution::can-handle-error-on-project) but could not determine a solution. Left comment.
- removed `solution-common` from `dependencies ` in `form` package - leads to package bloat

### DIFF
--- a/packages/common/src/resourceHelpers.ts
+++ b/packages/common/src/resourceHelpers.ts
@@ -59,20 +59,20 @@ import {
   IFileMimeType,
   IItemTemplate,
   IItemUpdate,
-  IMimeTypes,
   ISourceFileCopyPath,
   IUpdateItemResponse,
   UserSession
 } from "./interfaces";
 import { new_File } from "./polyfills";
 import {
-  addItemResource,
   updateGroup,
   updateItem,
   updateItemInfo,
   updateItemResource
 } from "@esri/arcgis-rest-portal";
-import { ArcGISAuthError } from "@esri/arcgis-rest-request";
+import { addResourceFromBlob } from "./resources/add-resource-from-blob";
+import { copyResource } from "./resources/copy-resource";
+
 import { updateItem as helpersUpdateItem } from "./restHelpers";
 import { getBlob, getBlobAsFile, getItemResources } from "./restHelpersGet";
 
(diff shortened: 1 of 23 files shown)


packages/common/src/resourceHelpers.ts
```diff
@@ -83,38 +108,38 @@ export function addMetadataFromBlob(
...
-    params: {}
-  };
-  if (folder) {
-    addRsrcOptions.params = {
-      resourcesPrefix: folder
-    };
-  }
-  return addItemResource(addRsrcOptions);
-}
+// export function addResourceFromBlob(
```
rweber-esri, line 111: Left intentionally?

rweber-esri:
@dbouwman 

LGTM. Just as a heads-up, there's a regression on this branch related to Feature Service post processing, which was introduced in [this commit](https://github.com/Esri/solution.js/commit/76bec4ab071a8edf706c31d83576410e1a3ecf65#diff-266ab901f19c8c388edb8eb5976c2c7b), which I've patched on [my pr](https://github.com/Esri/solution.js/pull/356/commits/f1a16a222db926ce8a79e41e6e4c6f28e29d39f9). I was able to successfully deploy a Solutions.js Survey Solution after applying the same fix to this branch, noticed no regressions in resource copying.

PR #364, opened 2020-06-05
Add migration to update survey map questions to use org default base …

rweber-esri:
…map then remove related web map templates.

When Hub templatizes a Survey that contains map questions, it creates a dependency/template for the base map. This PR:

1. Updates the solution migration process to support promises (but still serially processes them). 
2. Adds a migration to update any Form map questions to use the default org base map and removes their related Web Map templates.


rweber-esri:
@MikeTschudi 

I suspect this should point/merge to the [RC branch](https://github.com/Esri/solution.js/tree/release-candidate) instead of `develop`?

MikeTschudi:
No, please don't.  The RC branch is a temporary waypoint between `develop` and `master`--a chance to do a quick test of the compatibility between the solution-deployment-app and the solution.js library on servers before creating a point release.

Once the test is done, we'll create the point release in `master`, and merge it back into `develop`, preferably after this PR has been merged into `develop`. The RC branch is then deleted.

rweber-esri:
> No, please don't. The RC branch is a temporary waypoint between `develop` and `master`--a chance to do a quick test of the compatibility between the solution-deployment-app and the solution.js library on servers before creating a point release.
> 
> Once the test is done, we'll create the point release in `master`, and merge it back into `develop`, preferably after this PR has been merged into `develop`. The RC branch is then deleted.

Ok, thanks for the clarification.

packages/common/src/migrations/upgrade-two-dot-six.ts
```diff
@@ -0,0 +1,109 @@
...
+    );
+    clone.item.properties.schemaVersion = 2.6;
+
+    if (!formTemplates.length) {
+      return Promise.resolve(clone);
+    }
+
+    // fetch org details
+    return getPortal("", authentication)
+      .then(portal => {
```
dbouwman, line 50: This seems like stuff we should be doing as part of the deployment process as it is customizing the template for the current user's org. So if we created a template via Solution.js, it gets schemaVersion 3... how does that have the basemap stuff for the deploying user vs the person who created the template?

MikeTschudi:
Travis CI checks [passed](https://travis-ci.org/github/Esri/solution.js/builds/695143848) but Travis didn't update this PR.

packages/common/src/migrations/upgrade-two-dot-six.ts
rweber-esri: > This seems like stuff we should be doing as part of the deployment process as it is customizing the template for the current user's org. So if we created a template via Solution.js, it gets schemaVersion 3... how does that have the basemap stuff for the deploying user vs the person who created the template?

Form templates with map questions created by Solutions.js (schemaVersion 3), reference the same base map item (which is an esri-owned, publicly-accessible base map) as the source Form item & do not create a Web Map template for the basemap.

Form templates with map questions created by Hub create a separate Web Map template for the basemap, which @brollison wanted to avoid, so the decision was made to filter out the basemap templates and default the map questions to use the org default.

In hindsight, I agree, the actual overriding of the basemap probably makes more sense downstream in the type specific processing. Will refactor this PR to:

1) Migration will remain, but it will only filter the webmap templates
2) Org default basemap will be overridden downstream in the type specific processor

rweber-esri: > 1. Migration will remain, but it will only filter the webmap templates
> 2. Org default basemap will be overridden downstream in the type specific processor

@dbouwman @MikeTschudi 

Updated per my above comment, ready for re-review

packages/common/src/restHelpersGet.ts
```diff
@@ -65,7 +66,29 @@ export function getPortal(
   const requestOptions = {
     authentication: authentication
   };
-  return portalGetPortal(id, requestOptions);
+  return portalGetPortal(id, requestOptions).then(portalResponse => {
```
dbouwman, line 69: Why are we adding more logic to this centralized function that is simply called `getPortal`? It looks like this is getting some additional information (defaultBasemap?) so maybe put this in a second fn `getDefaultBasemapForPortal(portal, auth)` sort of a thing? I'm also not a fan of decorating well-known-structures (the portal response object) with new props as it becomes confusing as to what is coming from where. Also... pls add comments :)

dbouwman: And if when you do make a new fn, put it in it's own file

packages/deployer/test/deployer.test.ts
```diff
@@ -332,6 +343,16 @@ describe("Module `deployer`", () => {
               "/portals/abCDefG123456?f=json&token=fake-token",
             portalsSelfResponse
           )
+          .get(
```
dbouwman, line 346: Needing to make these sort of changes is indicative that the change could be more focused... i.e. did all these functions actually _need_ the basemap? if not, why are we making two additional xhrs in all these places?

packages/common/src/restHelpersGet.ts
rweber-esri: Ya, admittedly this is something I went back and forth on. The `getPortal` function is only called once, when we bootstrap the org details in the templateDictionary. I thought there _could_ be other opportunities to leverage this data, so opted to expand this method to return one additional property (but required 2 addl XHRs). I'm happy to move this out to it's own function though.

rweber-esri:
> Please refactor out the changes you made to `getPortal`, add a new function `getOrgDefaultBaseMap(portalSelf, auth)` (or whatever you want to call it) and use it where you specifically need this information.

@dbouwman I've pushed the requested changes per above.

PR #381, opened 2020-06-17
Use ArcGIS-Html-Sanitizer v2.1.1

dbouwman:
Back in February, `arcgis-html-sanitizer` [changed it's build system](https://github.com/Esri/arcgis-html-sanitizer/pull/51) to output ES6 modules instead of ES5 modules.

This has some upsides when all consumers of the library are able to drop IE11 support, but most packages will ship ES5 modules, not ES6.

While the Solutions App itself does not need to support IE11, Hub currently does. In order to bring Solution.js into Hub, we need to use the 2.1.1 version, which is what this PR does.

It also uses ES5 syntax in `index.html` in the `/demos/deploySolution` app so that it can run in IE11 (and is how I tested this)

### DIFF
--- a/demos/addResource/package-lock.json
+++ b/demos/addResource/package-lock.json
@@ -88,9 +88,9 @@
       }
     },
     "@esri/solution-common": {
-      "version": "0.16.6",
-      "resolved": "https://registry.npmjs.org/@esri/solution-common/-/solution-common-0.16.6.tgz",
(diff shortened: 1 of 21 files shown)


MikeTschudi:
[Travis CI](https://travis-ci.org/github/Esri/solution.js/pull_requests) appears to be unavailable right now.

On a separate computer, I ran unit tests and coverage, so going ahead with merge.

Issue #391, opened 2020-06-30
Cannot create solutions which contain a living atlas layer and a geocode service 


shoe913:
Web App Builder apps which contain a widget that uses a geocode service (search, screening, etc.) and a map with a living atlas layer cannot be converted to a solution item.

shoe913:
@jmhauck 

Is there anyway we can work to get this included into a build a little sooner rather than later? Several solutions we plan on releasing in July require this to build and we'd like to make sure we don't hit any other issues during testing.

chris-fox:
@shoe913, could you put together a simple repro item for @jmhauck? Would help him debug and get a quicker fix.

shoe913:
@jmhauck you can repro the error with this app. 


https://arcgis4localgov2.maps.arcgis.com/home/group.html?id=6c1e20ee307641b884c63e3b4545f2dc

chris-fox:
@shoe913 , could you please verify for 8.3. Thanks.

shoe913:
Veerified

PR #393, opened 2020-07-06
add fallback for getDefaultLocatorURL

jmhauck:
#391

### DIFF
--- a/packages/common/test/templatization.test.ts
+++ b/packages/common/test/templatization.test.ts
@@ -185,6 +185,33 @@ describe("Module `templatization`: common functions involving the adlib library"
       );
       expect(actual).toEqual(expected);
     });
+
+    it("should retain template variables when they are not in the templateDictionary", () => {
+      const template: any = {
+        item: {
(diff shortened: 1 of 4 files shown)


packages/common/test/templatization.test.ts
```diff
@@ -185,6 +185,33 @@ describe("Module `templatization`: common functions involving the adlib library"
       );
       expect(actual).toEqual(expected);
     });
+
+    it("should retain template variables when they are not in the templateDictionary", () => {
```
MikeTschudi, line 189: With the adlib library admonition that [Transform functions should be VERY resilient - we recommend unit testing them extensively](https://github.com/Esri/adlib#notes-about-transforms), perhaps add test(s) to be sure that you get `{{${GEOCODE_SERVER_NAME}}}` when expected?

jmhauck: @MikeTschudi the settings hash has no entry for any of these. The transform function no longer fails because of this and does return ` {{organization.helperServices.geocode:getDefaultLocatorURL}}` as now expected.

PR #397, opened 2020-07-09
Creator refactor for resource schema

dbouwman:
Adds migrations (disabled for now) to convert resource paths --> ISolutionResource objects

Also refactors some functions in `creator` package to do a few things:
- flatten promise chains
- extract logic from promise chains into more testable functions
- import only the functions needed vs `* as someName`

**Note**: There is code in here that's commented out - that is intentional as this does not "land" the conversion to `ISolutionResource` objects, but I wanted to get the refactors merged to avoid gnarly rebases if/when others touch the same areas of `creator`

**Note Part Deux**: In Hub.js I've added logic that hides functions marked w/ `@private` from the docs. I will also be adding that to this library so feel free to start using that.


packages/creator/src/creator.ts
```diff
@@ -35,228 +53,90 @@ import * as createItemTemplate from "./createItemTemplate";
...
-              groupInfo.thumbnail,
-              true
-            );
-          }
-
-          // Create a solution with the group contents
-          return groupItems;
-        },
+  let createOptions: ICreateSolutionOptions = options || {};
+  const progressCb = createOptions.progressCallback || noOp;
```
dbouwman, line 60: We could consider doing something like this in a number of places as it removes a whole bunch of checks to see if the callback exists... it either does, or it's a `noOp` function and either case, we just call it as needed

Issue #401, opened 2020-07-17
Solutionid type keyword not long enough


shoe913:
The type keyword created for solutionid needs to be 32 characters. We're currently only getting 8

Item:
https://localdeployment.maps.arcgis.com/home/item.html?id=2e9600ba0a4b46e8bbe6cd1849481f2c

Keywords:
`['Solution', 'solutionid-90zlov5f', 'solutionversion-1.0', 'Template']`

chris-fox:
verified

PR #403, opened 2020-07-17
Returned solutionid to a 32-char string

MikeTschudi:
#401


MikeTschudi:
I don't know the reason for the 32-char request, but it appears to be useful for the solution-creation process.

PR #405, opened 2020-07-22
Allow older hub Solution Template items

dbouwman:
Before the `Solution` item type was available Hub used `Web Mapping Application` with the `hubSolutionTemplate` keyword. Turns out there are a lot of these in PROD, so we need to allow them into the deployer.

This just opens up `isSolutionTemplateItem` to allow these through.


MikeTschudi:
Travis CI [passed](https://travis-ci.org/github/Esri/solution.js/builds/710906421) but is not notifying PR

PR #408, opened 2020-07-29
Simplify packages

MikeTschudi:
Removed peerDependencies: not needed and require manual updating during release builds


dbouwman:
@MikeTschudi I'm pretty sure we need `peerDependencies` in the actual packages. We can remove them from the `/demos` stuff, but `npm` uses `peerDependencies` to inform a user of the other things a package depends on... at least in the near term we should keep those.

PR #425, opened 2020-08-13
Added a dependency graph and some error checking

MikeTschudi:
* Updated solution.js library to report items involved in a cycle
* Updated getItemInfo demo to report duplicate items in a template, to report items involved in a cycle, and to draw the dependency graph for a Solution
* Patched npm
* Added test for 100% coverage
* Changed names of external functions to reflect their externality


demos/getItemInfo/index.html
```diff
@@ -85,11 +88,28 @@
               username: document.getElementById("username").value,
               password: document.getElementById("password").value,
               portal: document.getElementById("srcPortal").value + "/sharing/rest"
-            })
+            }),
```
jmhauck, line 91: ;

MikeTschudi: Ugh--thank you for catching this

PR #435, opened 2020-08-19
add optional configOptions

jmhauck:
This should be the only change necessary to support the user config options from solution.js side. 

The solutions app will know how to dynamically render the UI to collect information from the user based on what is provided here. It will then feed the user defined options into the templateDictionary used for variable swapping by deploy.


MikeTschudi:
Why put this in the solution.js library, please? The library is independent of the app and shouldn't have app-specific content.

jmhauck:
Was just looking for a placeholder...I suppose I could switch the app to think that what comes in as ISolutionData is actually just "any" rather than use that interface and this wouldn't be necessary.

MikeTschudi:
Or create a new type in that app that inherits from ISolutionItemData...

PR #438, opened 2020-08-26
Enh/build processing

MikeTschudi:
* Updated hub version
* Removed source maps as part of regular build
* Updated TypeScript and tslib
* Standardized usage of devDependencies, peerDependencies, and dependencies

PR #445, opened 2020-09-03
Guard against non-templated items in Group deps

dbouwman:
When a group is included in the templating process, the system attempts to include all the items shared to the group. However, not all items can be templated - i.e. `Hub Initiative` items. 

This PR just adds a guard into the `_postProcessGroupDependencies` function so it does not attempt to process the dependantTemplate if it's not found in the templates array.

### DIFF
--- a/packages/creator/src/helpers/add-content-to-solution.ts
+++ b/packages/creator/src/helpers/add-content-to-solution.ts
@@ -28,7 +28,7 @@ import {
   SItemProgressStatus,
   updateItem
 } from "@esri/solution-common";
-import { getProp } from "@esri/hub-common";
+import { getProp, getWithDefault } from "@esri/hub-common";
 import { UserSession } from "@esri/arcgis-rest-auth";
 import {
   createItemTemplate,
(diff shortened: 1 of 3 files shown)


packages/creator/src/helpers/add-content-to-solution.ts
```diff
@@ -203,14 +203,22 @@ export function _postProcessGroupDependencies(
...
+            dependantTemplate,
+            "dependencies",
+            []
+          ).indexOf(id);
+          /* istanbul ignore else */
+          if (gIndex > -1) {
+            removeDependencies = true;
+          }
+          /* istanbul ignore else */
+          if (dependantTemplate.groups.indexOf(id) < 0) {
```
jmhauck, line 219: @dbouwman groups are always good in this type of case? and shouldn't do something similar like:

dependantTemplate?.groups.indexOf(id) < 0

jmhauck (approved):

packages/creator/src/helpers/add-content-to-solution.ts
dbouwman: Good question @jmhauck - in theory - `IItemTemplate` has a `.group` property, so it should be fine, but I'll add `dependantTemplate.groups?.indexOf(..)` just to cover us...

dbouwman: Ok - I swapped this around to be even more defensive... so if `.groups` does not exist, it will add it to the `dependantTemplate`... not sure this is possible in the wild, but... I'm a fan of things not being breakable :)

dbouwman:
@MikeTschudi Lets merge your #438 and then this one, and then can we publish a release of Solution.js?

dbouwman:
Something is wrong w/ the package.json and package-json.lock files - I started looking into it but have to head out for the weekend... we should hold off merging this for now.

PR #451, opened 2020-09-04
retain params that have already been defined in template dict

jmhauck:


jmhauck:
@MikeTschudi To support user configuration options from the deploy app we need to add a _params: {}_ to the solution item that contains the info we need to construct the UI the end user will interact with. After they choose the values we add them to base template dictionary that we pass to solution.js for the deployment. This makes sure that we honor the user defined options rather than the spec that was used for the UI construction.

MikeTschudi (approved):

Issue #458, opened 2020-09-10
Wrong authentication passed when fetching item template


mpayson:
[getItemMetadataAsFile](https://github.com/Esri/solution.js/blob/develop/packages/deployer/src/deployer.ts#L75) passes in the deployment recipients session (`authentication`) resulting in a `403` error -- this should be the storage session (`storageAuthentication`). cc @MikeTschudi

```javascript
// ...
return Promise.all([
  Promise.resolve(model.item),
  Promise.resolve(model.data),
  common.getItemMetadataAsFile(model.item.id, storageAuthentication)
]);
// ...
```

mpayson:
Confirmed this was fixed in the latest releases, closing -- thanks!

MikeTschudi:
Thank ***you***, Max!

PR #459, opened 2020-09-10
Update authentication for fetching item template

mpayson:
#458


MikeTschudi (approved):
Thank you for the catch!

PR #460, opened 2020-09-10
Updated arcgis-rest-js and hub.js libraries

MikeTschudi:
* arcgis-rest-js from 2.16.0 to 2.17.0
* hub.js from 6.4.0 to 6.9.1


MikeTschudi:
Switching to #461, which solves Travis package-lock problem

PR #475, opened 2020-09-17
remove lookbehind from id test regex

jmhauck:


MikeTschudi (approved):
I'm glad that you saved the regExTest expression; maybe Safari will catch up someday and we can use it!

PR #505, opened 2020-10-12
view spatial reference

jmhauck:


packages/deployer/src/deploySolutionItems.ts
```diff
@@ -577,8 +590,12 @@ export function _createItemFromTemplateWhenReady(
   // ensure this is present
   template.dependencies = template.dependencies || [];
   // if there is no entry in the templateDictionary, add it
-  if (!templateDictionary.hasOwnProperty(template.itemId)) {
-    templateDictionary[template.itemId] = {};
+  if (
+    !templateDictionary.hasOwnProperty(template.itemId) ||
+    !getProp(templateDictionary[template.itemId], "def")
```
MikeTschudi, line 595: "def" seems like a magic word--perhaps add an inline comment or reference about it?

packages/common/src/featureServiceHelpers.ts
```diff
@@ -855,6 +859,52 @@ export function _updateTemplateDictionaryFields(
...
+  if (spatialReference) {
+    setCreateProp(
+      templateDictionary,
+      `${itemId}.defaultSpatialReference`,
+      spatialReference
+    );
+  }
+}
+
+export function validateSpatialReference(
```
MikeTschudi, line 877: Please add some documentation to describe what these functions do

MikeTschudi (approved):

PR #512, opened 2020-10-14
Excluding site itemInfo from postprocessing and own children array

drewctate:


drewctate:
@MikeTschudi if this looks okay to you, it would be awesome if we could get a release in the next few days. Thanks so much!

dbouwman (approved):
Site-wise this looks good

MikeTschudi (approved):

PR #517, opened 2020-10-19
Tweak demo setup to fix `global` problem and symbolic links

MikeTschudi:
1. Recent package updates and changes to improve security may have led to a sporadic problem with `global` being defined for modules:
image
2. Symbolic links are not being set up for the demo apps, and so they are not taking advantage of pre-publication work in the packages.

For the first problem, added the line
```js
intro: 'var global = typeof self !== undefined ? self : this;',
```
to umd-base-profile.js to make sure that `global` is always set up.

For the second problem, removed `command.bootstrap.ignore` from lerna.json.

It also appears that one has to run `npm install` an extra time to get the symlinks to be generated:
```bat
echo ===== install top level =====
call npm install

echo ===== update examples =====
pushd demos
call npm install
call npm run build
popd

echo ===== set up symbolic links =====
call npm install
```

### DIFF
--- a/demos/package-lock.json
+++ b/demos/package-lock.json
@@ -1,385 +1,385 @@
 {
-  "name": "demos",
-  "version": "0.19.1",
-  "lockfileVersion": 1,
-  "requires": true,
-  "dependencies": {
-    "@esri/arcgis-html-sanitizer": {
-      "version": "2.1.1",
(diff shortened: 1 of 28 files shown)


MikeTschudi:
Moved to issue #520

Issue #457, opened 2020-09-09
Store the url property of items in the replacement dictionary to be available to other templates in the solution


chris-fox:
Currently we store the item id of as a replaceable variable so that it can be swizzled in other items. For example:

{{cd4e7edbca5c428a86e755e701a438b5.itemId}} 

This request would be to also add the url property of the item to the replacement dictionary:

{{cd4e7edbca5c428a86e755e701a438b5.url}}

This is necessary for items like web experiences because the url can change between Online and Enterprise. In my example, I had a Hub site referencing a web experience and there was no way to create a variable that could be replaced and be accurate for both Enterprise and Online.

MikeTschudi:
@chris-fox, would you please expand on this? Is it like #477, except that for each item with a URL (e.g., a Web Map), we add a template-dictionarly mapping from the [Web Map's] URL to {{*webmap_id*.url}} for replacing the URL in other items?

chris-fox:
@MikeTschudi, yes that's exactly right. Some items like Experience Builder apps have complex logic for resolving their url so it works in Online and Enterprise. So after that is resolved it would be good to store that url in the template-dictionary, so that any items that want to reference it and just use the {{itemid.url}} variable to resolve it.

chris-fox:
@MikeTschudi, been testing this and it doesn't seem to be working the way I would expect. The solution below has a web map that in its item description has the following:

{{fcb2bf2837a6404ebb418a1f805f976a.url}} <div>{{cefb7d787b8b4edb971efba758ee0c1e.url}}</div>

These are references to a Web AppBuilder and Experience Builder app items that are dependencies of the web map. When I deploy those variables are not being replaced with the url's of the apps even those items do have a url property.

https://www.arcgis.com/home/item.html?id=9e59550a94914f9b8a5e71814968bc35

MikeTschudi:
@chris-fox, got it. Implemented url addition only to solution creation as with #477. Will add to deployment. Thanks for the example!

chris-fox:
@MikeTschudi, that was going to be my next request to get this implemented in the post process workflow during creation as well. So that's great. I will test this out as well.

chris-fox:
Hi @MikeTschudi, I was testing this implementation of the item url property during solution creation with #477 and it is not working as I would expect. You can repro with the following group:

https://localdeployment.maps.arcgis.com/home/group.html?id=04f36eae6346438c873a461da54c4a2f

In the group is a web map and in its item description is the following text:

[...]

PR #525, opened 2020-10-26
Update URLs during deployment

MikeTschudi:
https://github.com/Esri/solution.js/issues/457#issuecomment-713594283


packages/web-experience/src/web-experience-processor.ts
```diff
@@ -125,19 +126,36 @@ export function createItemFromTemplate(
...
     });
 }
+
+// ------------------------------------------------------------------------------------------------------------------ //
+
+/**
+ * Flags a failure to create an item from a template.
+ * @return Empty creation response
+ */
+export function _generateEmptyCreationResponse(
```
jmhauck, line 152: @MikeTschudi why all the redundancy with this function across so many modules?

jmhauck (approved):

packages/web-experience/src/web-experience-processor.ts
MikeTschudi: You're right--I'll merge them into `common`. Thanks for pointing it out.

jmhauck (approved):

Issue #535, opened 2020-10-29
Error message from missing metadata in solution template sent to deployed solution item


MikeTschudi:
Visible in Enterprise with Firefox

ghudgins:
watching this for the 10.9 pre-release code freeze

chris-fox:
Verified

PR #537, opened 2020-10-29
Improved handling of missing Solution item metadata

MikeTschudi:
#535 

* Improved handling of missing Solution item metadata
* Don't fetch or upload metadata.xml for the solution item if the Solution's id is not a GUID (for Hub)
* Don't send the thumbnailUrl for the deployed Solution item if its `.thumbnail` property is undefined

### DIFF
--- a/packages/deployer/src/deployer.ts
+++ b/packages/deployer/src/deployer.ts
@@ -28,7 +28,7 @@ import {
   isSolutionTemplateItem,
   updateDeployOptions
 } from "./deployerUtils";
-import { IModel } from "@esri/hub-common";
+import { IModel, isGuid } from "@esri/hub-common";
 
 /**
  * Deploy a Solution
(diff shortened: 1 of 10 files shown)


packages/deployer/src/deployer.ts
```diff
@@ -68,11 +68,13 @@ export function deploySolution(
...
       } else {
-        // fetch the metadata and pass the item & data forward
+        // fetch the metadata if the model's id is a GUID and pass the item & data forward
         return Promise.all([
           Promise.resolve(model.item),
           Promise.resolve(model.data),
-          common.getItemMetadataAsFile(model.item.id, storageAuthentication)
+          isGuid(model.item.id)
+            ? common.getItemMetadataAsFile(model.item.id, storageAuthentication)
+            : Promise.resolve()
```
dbouwman, line 77: We're sure that the code further down can handle `itemMetadata` as `undefined`?

MikeTschudi: Yes--`deploySolutionFromTemplate` now checks if the metadata is defined before updating the deployed solution item with it:
```ts
// Pass metadata in via params because item property is serialized, which discards a File
const additionalParams: any = {};
/* istanbul ignore else */
if (solutionTemplateMetadata) {
  additionalParams.metadata = solutionTemplateMetadata;
}
return common.updateItem(
  solutionTemplateBase,
  authentication,
  deployedFolderId,
  additionalParams
);
```

dbouwman (approved):
Thanks for jumping on this!

PR #554, opened 2020-11-30
B/156515 handle surveys with zip files named undefined

rweber-esri:
[Target Process: 156515](https://esriarlington.tpondemand.com/entity/156515-older-deployed-survey-templates-missing-zip)

Surveys created before Hub switched to using the Survey123 API can have the literal string `"undefined"` as the item name which results in an error when attempting to create a solution template. This PR refactors the resource helpers `storeFormItemFiles` method to account for this.

### DIFF
--- a/packages/common/src/resourceHelpers.ts
+++ b/packages/common/src/resourceHelpers.ts
@@ -774,8 +774,12 @@ export function storeFormItemFiles(
 
     // Store form data
     if (itemData) {
+      const originalFilename =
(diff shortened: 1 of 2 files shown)


jmhauck (approved):

rweber-esri:
@MikeTschudi 

Are we currently waiting for any other PRs? If not, would it be possible to cut a patch release so we could pull this fix into Hub?

PR #568, opened 2020-12-28
Changed source of type for compatibility

MikeTschudi:
Even though two paths lead to same base type, TypeScript generates
```
Types have separate declarations of a private property '_user'.
```


MikeTschudi:
Analyze and CodeQL appear to be broken; will retry

PR #570, opened 2020-12-30
Updated package.json versions

MikeTschudi:
package.json updates:

* Got rid of adlib from demos because it's brought in via rest of library
* Updated tslib

And package-lock.jsons as a result.


MikeTschudi:
Jasmine's spyOn and tslib 2.0.3 appear to be incompatible:  "Error: <spyOn> : ___ is not declared writable or has no setter"

PR #573, opened 2021-01-07
166968 - Fix intermittent Hub survey template deployment failures.

rweber-esri:
[166968](https://esriarlington.tpondemand.com/entity/166968-regression-survey-solution-templates-fail-to).

We're experiencing intermittent deployment failures for Hub survey templates. When the error occurs, the message is `Item does not exist or is inaccessible` and is related to the logic that moves the `Form` & `Feature Service` items and deletes their original folder created by the S123 API. Despite the  REST `move` API returning a success response for both items, the `Form` remains in the S123 folder and gets deleted when the original S123 folder gets deleted. The actual error is thrown from a downstream `update` call for the `Form` in the post processing logic.

This PR resolves in the intermittent failures by moving the calls that move the items & delete the original S123 folder from `create-item-from-template` to the end of the `post-processing` logic.

### DIFF
--- a/packages/form/src/helpers/create-item-from-hub-template.ts
+++ b/packages/form/src/helpers/create-item-from-hub-template.ts
@@ -20,14 +20,12 @@ import {
   IItemProgressCallback,
   ICreateItemFromTemplateResponse,
   EItemProgressStatus,
-  removeFolder,
   replaceInTemplate,
   updateItemExtended,
   ISurvey123CreateParams,
   ISurvey123CreateResult,
   getItemBase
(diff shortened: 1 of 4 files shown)


MikeTschudi (approved):

jmhauck (approved):

rweber-esri:
@MikeTschudi @jmhauck 

Thanks for the review. Not sure if we're holding for anything, I see the latest release is at `v0.21.3` but patch tags `v0.21.4` & `v0.21.5` already exist. If not, would it be possible to merge and publish a new patch version?

Issue #587, opened 2021-01-26
Need to handle adding layers to views with source of itself need to be made in separate addToDefinition call


chris-fox:
It's possible to create views where one layer has a source coming from another service and another layer in the view has a source of the first layer in the view. 

In the solution item below there is an example. The template id 'af0451ace705472cbe3a4306205a53e4' has 2 layers. The first layer is a multisource join view on top of the feature service with the id '589c0147e412497fad70d7074ed13261'. The second layer is a view that has a source of the first layer in the same service id 'af0451ace705472cbe3a4306205a53e4'.

Adding both layers in a single addToDefinition call create an invalid second layer and fails the deployment. If we add each in separate calls, it successfully creates both layers.

https://www.arcgis.com/home/item.html?id=6b31469f128d46008b790c0d6b37b44a

jmhauck:
#588

chris-fox:
Verified

PR #588, opened 2021-01-28
split add to def further

jmhauck:


MikeTschudi (approved):
Tidy solution of tricky problem!

jmhauck:
#587

PR #597, opened 2021-02-17
Added option to enter solution id in copySolutions

MikeTschudi:
In addition to list of solutions in source organization, an option is provided to enter the id of a solution item:
image


chris-fox (approved):
good change to the demo

Issue #598, opened 2021-02-18
Stack overflow: solution.js doesn't play well with Angular Zone.js


xiao8579:
**Describe the bug**
When deploying the Pipeline Hazards Patrol solution in Designer, the deployment process hangs at `addItem` call for the Web Experience item. At this point, all the rest items inside this solution have been created already, only missing the web experience item in the target folder.

This is a stack overflow issue and happens on both Production and DEVEXT versions of designer.
Current solution.js version in Designer:  0.20.1

**To Reproduce**
1. Go to https://devext.arcgis.com/apps/quickcapture/projects (QuickCapture_Web | QuickCapture1)
2. Click **New project**, select **Start from template**
3. Click **From organization**.
4. Select **Pipeline Hazards Patrol** and click **Use template** 
5. See the error below.
![image](https://user-images.githubusercontent.com/18433455/108418227-cd417280-71e5-11eb-84bd-67f037303d2e.png)

**Additional context**
1. The issue seems to only affect the QuickCapture Designer client so far, but it stops the product team from creating a QuickCapture solution template that will include a web experience app, and even it's created, the deployment process will fail for our users. Survey123 Web (based on Agular too) is also planning to adopt solution.js, this issue will potentially affect S123 in future.
2. The issue shows that Solution.js doesn't play well with Zone.js. QuickCapture Designer can't stop using zone.js because it's a core dependency of Angular. 
2. Based on my testing, the deployment process will fail in Designer for all Web Experience when included in the solution item.

Cc: @MikeTschudi

MikeTschudi:
@xiao8579, would you please provide an Angular test case that doesn't involve the QuickCapture Designer?  From what I can tell using the Designer, the compressed code doesn't match solution.js, and so it's not possible to understand or debug what is happening.

jonpaez:
@MikeTschudi, here are some steps to create an angular app:

1. Install the angular cli
`npm install -g @angular/cli`

2. Navigate to the folder you want to create your app in and run
``` bash
ng new solution-test
cd solution-test
[...]

PR #599, opened 2021-02-22
Modified item handling to support Angular

MikeTschudi:
The deployment code for four item types--Web Experience, Hub Site Application, Hub Page, and StoryMap--isolated the routines from side effects to make the routines more robust and testable.

This turns out to be a problem when embedding the code into Angular, which has its own definition for a Promise--a definition that contains circular references and thus breaks during certain JavaScript options.

This PR changes the four item types to allow side effects. Not an enhancement, but it does permit Angular to host solution.js.

#598


dbouwman (approved):
Looks great! Thanks @MikeTschudi

Issue #593, opened 2021-02-16
Sometimes items are returned in search that no longer exist causing failure to deploy.


chris-fox:
I have seen this in Enterprise a few times. When we deploy and we search for existing items a result is returned, I can see this in the UI as well:

![image](https://user-images.githubusercontent.com/6796763/108077262-5156f700-703a-11eb-8ec2-d267a5ef56da.png)

However, when we go attempt to get the item it no longer exists:

![image](https://user-images.githubusercontent.com/6796763/108077361-70558900-703a-11eb-97c1-05de60d61761.png)

![image](https://user-images.githubusercontent.com/6796763/108077422-806d6880-703a-11eb-8cc8-edb6802de3c6.png)

This causes the deployment to fail. It seems like there is an issue with the search index.

I would like to handle this case and if an item returned by the search fails when we attempt to fetch it we should just treat it as if the item was not found and ignore it.

PR #600, opened 2021-02-23
stale search index issue

jmhauck:
This also adds an additional set of requests to obtain the fields for layers and tables when we are leveraging existing item reuse. Portal will fail when a view has field references that don't exist in the source service. We fixed this issue for "standard" deployment but the same issue can occur with existing items.

#593


MikeTschudi (approved):
This looks like it was a hairy challenge!

PR #601, opened 2021-02-25
Refactor storeFormItemFiles to degrade to itemId.zip vs formData.zip …

rweber-esri:
…to avoid collisions if multiple surveys in the solution must degrade.

[12](https://devtopia.esri.com/dc/hub/issues/12)

This is related to [this fix](https://github.com/Esri/solution.js/pull/554). We currently run into a collision which fails deployment when multiple surveys in the Solution must degrade to the fallback zip file name. This fix changes the fallback zip name to `${itemTemplate.itemId}.zip` vs `formData.zip` to avoid the collision when copying these assets to the folder.


jmhauck (approved):

rweber-esri:
Thanks @jmhauck & @MikeTschudi.

Would is be possible to publish a patch release v0.22.2 so Hub can pull in this fix?

Issue #604, opened 2021-03-10
Multiple update calls made for CSV item types, sometimes resulting in a failure


chris-fox:
I was looking into a random failure in Enterprise for the following item and found a random issue where the following solution would fail to deploy: 0f3f8a880370441f914732154f24c42d

In the solution there is a template c2f46d1a16234cef95f120c0cfecedd2 that sometimes fails on update the thumbnail. There are 2 back to back updates on that item and we got a 500 error back on the 2nd update. 

I remember we had issues in the past where we had back to back updates on items. I think we changed the logic by setting the thumbnail in the original addItem call to reduce the number of calls. I am wondering if we could do the same here

chris-fox:
This is looking great Mike.

PR #605, opened 2021-03-15
Centralize setting of thumbnail during deployment

MikeTschudi:
See issue #604


packages/common/src/resourceHelpers.ts
```diff
@@ -754,6 +723,33 @@ export function isSupportedFileType(filename: string): boolean {
   );
 }
 
+/**
+ * Gets the thumbnail of an item or group.
+ *
+ * @param storageAuthentication Credentials for the request to the storage
```
jmhauck, line 729: Could you change to "authentication" please

jmhauck (changes requested):

packages/common/src/resourceHelpers.ts
MikeTschudi: So that the doc matches the code? Why would I want that?  ;-)
Thanks for the catch

jmhauck (approved):

Issue #611, opened 2021-03-24
Resources not always added to item during deployment


chris-fox:
Occasionally during deployment resource files aren't being added to the item. When watching in the debugger all addResource request return successful, but when you inspect the item after not all the resources were added.

I can consistently reproduce with the solution below in ArcGIS Enterprise, and fairly regularly reproduce in ArcGIS Online: 

aeb7fe7d547c4ebda9ecddc829980550

The solution has a Quick Capture template (3b124a729ac24fceb0c25f7de3b1f11) Rapid Wildlife Reporter that has 22 svg resources used in the UI of the Quick Capture project. Often 1 or more is not added to the item during deployment. 

It seems like a bug in the REST api, but wondering if there is a way we can workaround it and attempt to re-add if the resource isn't actually added to the item.

chris-fox:
Verified

PR #615, opened 2021-03-25
Combine updates of resources into a single zip file during deployment

MikeTschudi:
#611

Note that there's an AGO-imposed limit of 50 resource files per zip file, which this modification does ***not*** handle; see https://github.com/Esri/solution.js/pull/615#discussion_r601510049.


MikeTschudi:
@dbouwman, doesn't appear that CodeQL is working right now

packages/common/src/resourceHelpers.ts
```diff
@@ -226,95 +231,110 @@ export function copyFilesFromStorageItem(
...
+    // many updates in a row to the same item: it claims success despite randomly failing
+    const resourceFilePaths = filePaths.filter(
+      filePath => filePath.type === EFileType.Resource
+    );
+    let zip: any;
+    if (resourceFilePaths.length > 0) {
+      zip = new JSZip();
+      awaitAllItems = awaitAllItems.concat(
+        // Note that AGO imposes a limit of 50 files, which is not checked in this code
+        // https://developers.arcgis.com/rest/users-groups-and-items/add-resources.htm
```
dbouwman, line 292: if/when you tackle the 50 file limit, hub.js has `chunkArray` (https://esri.github.io/hub.js/api/common/chunkArray/) which seems like something that could be useful here...

dbouwman:
Have you looked at using `batch` from hub.js? Asking because that's what I'm using in templates.js, and has worked reliably in my testing against the api... I believe my last bulk test showed that a concurrency of 2 was reliable, and that's what the default is in templates.js. That said, I think you have much more real-world experience around this, so if zipping is _the only reliable option_ we should consider how to address this in templates.js (where we _really_ want to avoid pulling in that big zip library)

MikeTschudi:
I'm interested in trying `batch`. We were already pulling in the jszip library, and the AGO 9.1 release is very close, so I went with it.

In this particular case, the files that had a dropping problem averaged 6 KB: lots of small files rapidly sent.

Issue #617, opened 2021-03-25
A user can belong to a maximum of 512 groups


MikeTschudi:
@dbouwman discovered that deployment fails if the user already belongs to the maximum number of groups.

Hub site creation checks for this, but a recent modification to this repo ([Solution deployment using groups not owned by user](https://github.com/Esri/solution.js/issues/390)) changes the nominal number of groups to ones that the user *owns*, not the number that he/she *belongs to*. As a result, Hub site creation is checking a false number and fails. The Hub team has a user who has encountered this.

But this is a general failure point for deployment.

@chris-fox, for 9.1?

chris-fox:
@MikeTschudi, what is the fix/change in this case?

MikeTschudi:
If the number of groups that the user belongs to + the number of groups in the solution + a group for the deployment > 512, provide an indication that deployment is not possible.

chris-fox:
I would say no for 9.1 then, because we would need to work on the UI/strings for how we communicate it to the user.

dbouwman:
fwiw - I think we've hit this with one actual customer, ever. 

We do run into this when our e2e test accounts, and some of our interactive testing accounts, accrue >511 groups over time... that is how we originally found this issue, and how it cropped up again earlier today.

MikeTschudi:
Suggestion from Dave is to filter user-owned groups into new template property `user.ownedGroups` rather than modifying `user.groups`; this will make the code clearer and avoid surprises such as he found.

PR #618, opened 2021-03-25
Fetch user when creating IHubRequestOptions

dbouwman:
The hub-types package had been pulling the user from `templateDictionary.user` and using that to construct the `IHubRequestOptions` hash that's needed by the hub-js functions that create sites etc.
However, the groups attached to that user are currently filtered to only the groups the user owns. This decreses the count of groups, which is a problem when a user is a member of 512 groups... the Hub code checks how many groups they belong to (via user.groups), and since it's filtered, it returns a number that's lower and then it tries to create more groups... and the api calls fail... and things go badly.

In this change, we always fetch the user from the platform, and use that instead.

### DIFF
--- a/demos/package-lock.json
+++ b/demos/package-lock.json
@@ -95,16 +95,13 @@
         "tslib": "^1.13.0"
       }
     },
-<<<<<<< HEAD
     "@esri/solution-common": {
-      "version": "0.22.5",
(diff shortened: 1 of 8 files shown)


MikeTschudi:
See also #617

MikeTschudi (approved):

Issue #630, opened 2021-05-11
Workforce Items no longer appear in item tree


shoe913:
When creating a solution item with workforce items, the workforce project no longer appears in the item tree. 

Sample Solution Item: https://arcgis4localgov2.maps.arcgis.com/home/item.html?id=07568ae3f6894c1eb564d249ee71f8cc 
Sample Group: https://arcgis4localgov2.maps.arcgis.com/home/group.html?id=eda80b5a07ff4457b515447e0173b7f3#overview

MikeMillerGIS:
@jmhauck would this fix be available in the solutionstest app.  @rtharrington still cannot create a solution from the electric outage.  We are stuck with testing till this is resolved.

jmhauck:
@MikeMillerGIS or @rtharrington could I have a link to the item? Also what app? I tested with the deploySolution demo app from this repo but could take a look at whatever app you are using.

MikeMillerGIS:
User: jhauck_arcgissolutions 
Group: https://arcgissolutions.maps.arcgis.com/home/group.html?id=7b77116d00824b6f923e5a456a30030a#overview

Create a solution from this group, the content do not show from the resulting item.

@rtharrington what was the link to the app used to create the solution

jmhauck:
Thanks...I'll take a look. And my bad...I said deploy but the fix was on the create side.

rtharrington:
> Thanks...I'll take a look. And my bad...I said deploy but the fix was on the create side.

FYI...When I created the solution, the solution items doesn't list any of the content like it usually does. 

![image](https://user-images.githubusercontent.com/25378341/118685186-dfc14d00-b7d0-11eb-97d3-8018035bcff5.png)

jmhauck:
@rtharrington what was the link to the app used to create the solution? I am seeing content listed when I use the createSolution demo app from this repo.

![image](https://user-images.githubusercontent.com/2170893/118687675-8520f480-b7c2-11eb-8e35-a7b562ef83d2.png)

rtharrington:
> @rtharrington what was the link to the app used to create the solution? I am seeing content listed when I use the createSolution demo app from this repo.
> 
> ![image](https://user-images.githubusercontent.com/2170893/118687675-8520f480-b7c2-11eb-8e35-a7b562ef83d2.png)

I used https://solutionstest.esri.com/create/index.html?sortField=modified&sortOrder=desc#create

jmhauck:
[...]

PR #631, opened 2021-05-12
dependency issue

jmhauck:
#630

_postProcessGroupDependencies and postProcessWorkforceTemplates both can intentionally remove items from the dependency array.


MikeTschudi (approved):
Interesting interaction!

PR #633, opened 2021-05-13
Delete version 0 deployed Solution items and the items deployed with them

MikeTschudi:
These are deployed Solution items before the `version` property was added to the `metadata` section


jmhauck:
Nice!

jmhauck (approved):

PR #645, opened 2021-05-28
Disable update def calls

jmhauck:
As we are working through issues with FS folks...testing to see where we may land if all updateDef calls can be removed.


chris-fox:
@jmhauck, can we not merge this into release as it is just a test?

PR #658, opened 2021-06-08
async if sync fails and remove update def for online

jmhauck:


MikeTschudi (changes requested):
Some new functions in featureServiceHelpers & restHelpers are not documented, but please wait; I'm trying to merge `develop` into the feature branch, and it's not yet working because hub.js requires an older version of arcgis-rest-js than we can use. Working with hub team to try to get hub.js updates.

MikeTschudi (approved):

Issue #679, opened 2021-06-28
Many small zips to the same item sometimes fail


MikeTschudi:
Creating a solution out of group 2efb053b7c424792ac9b611c3a32875f can fail because there are many sources supplying resources, and so lots of small zips (47 1-file zips, one each of 3-, 4-, 7-file zips, and 2 41-file zips) are being sent to AGO, which can't handle lots of small zips sent rapidly to the same item just like it can't handle lots of resource files sent rapidly to the same item. The resources logic needs to be changed to queue all of the resources for the Solution item and then send them in one set of zip files (possibly more than one zip file in set because of the AGO 50-file limit per zip).

The existing code also stores a link from an item to its resource in the solution item before knowing that that resource even exists (e.g., metadata files). The fix is to fetch the resource first and only store the link if the resource exists, and then save the resource for the mass zip at the end of solution creation.

See https://github.com/Esri/solution.js/issues/652#issuecomment-868884676

chris-fox:
@MikeTschudi, I tested some this morning and tried packaging 2efb053b7c424792ac9b611c3a32875f 2x into a new Solution, each time a resource failed to be included in the resulting solution item which caused it to fail when deployed. It failed for different items each time. 

The second time I tried I got the following error in the create demo app:

![image](https://user-images.githubusercontent.com/6796763/125318711-1a92bd80-e308-11eb-9247-2197eb7d3214.png)

I did verify that we are properly skipping living atlas layers in the latest fix.

PR #682, opened 2021-06-30
Bundle resource sending to created Solution item

MikeTschudi:
#679


packages/common/src/resources/copyAssociatedFiles.ts
```diff
@@ -197,3 +237,59 @@ export function copyAssociatedFiles(
     }
   });
 }
+
+/**
+ * Copies one or more zipfiles to a storage item.
+ *
+ * @param zipInfos List of zip files containing files to store
+ * @param sourceAuthentication Credentials for the request to the source
```
jmhauck, line 245: not used

jmhauck (approved):

PR #688, opened 2021-07-08
Fix resource handling during creation to skip non-existent files

MikeTschudi:
* Changed sending of resources to a Solution during Solution creation. Previously, only the resources from a single item were zipped. Now, resources are accumulated from all items and are zipped together.
* Fetch resources from source items to make sure that they exist before adding a reference to them in the created Solution. Previously, a reference to an item's resource was added to the item's template and fetching occurred later--when the item's resources were to be fetched, zipped, and sent to the created Solution. Now, the resource file itself is collected when the item's template is created (not just the path to it) and only zipping and sending occur at the end of the creation process.

see issue #679

### DIFF
--- a/packages/common/src/resources/copyResourceIntoZip.ts
+++ b/packages/common/src/resources/copyResourceIntoZip.ts
@@ -30,7 +30,6 @@ import { getBlobAsFile } from "../restHelpersGet";
  * Copies a resource into a zipfile.
  *
  * @param file Information about the source and destination of the file such as its URL, folder, filename
- * @param sourceAuthentication Credentials for the request to the source
  * @param zipInfo Information about a zipfile such as its name and its zip object
  * @return The result of the copy
  */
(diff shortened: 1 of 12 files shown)


packages/common/src/resources/copyResourceIntoZip.ts
```diff
@@ -25,6 +26,30 @@ import { getBlobAsFile } from "../restHelpersGet";
 
 // ------------------------------------------------------------------------------------------------------------------ //
 
+/**
+ * Copies a resource into a zipfile.
+ *
+ * @param file Information about the source and destination of the file such as its URL, folder, filename
+ * @param sourceAuthentication Credentials for the request to the source
```
jmhauck, line 33: source auth not used

packages/common/src/resources/getItemResourcesFilesFromPaths.ts
```diff
@@ -0,0 +1,54 @@
...
+
+/**
+ * Fetches a set of resources defined by paths.
+ *
+ * @param resourceItemFilePaths Paths to resources in source
+ * @return A promise which resolves with an array of resource files
+ */
+export function getItemResourcesFilesFromPaths(
+  resourceItemFilePaths: ISourceFileCopyPath[],
+  authentication: UserSession
```
jmhauck, line 30: auth missing in doc

jmhauck (dismissed):

PR #697, opened 2021-07-30
add custom params to deploy demo

jmhauck:
Allows us to pass custom params like wkid or the upcoming feature service id.

### DIFF
--- a/demos/src/deploy-solution-main.ts
+++ b/demos/src/deploy-solution-main.ts
@@ -120,9 +120,9 @@ export function deploySolution(
     consoleProgress: true,
     storageAuthentication: srcAuthentication,
     enableItemReuse,
-    templateDictionary: {
+    templateDictionary: isJsonStr(customParams) ? {
       params: JSON.parse(customParams)
-    }
+    } : {}
   };
   const itemUrlPrefix = destAuthentication.portal.replace("/sharing/rest", "");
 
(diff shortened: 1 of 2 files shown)


demos/src/deploy-solution-main.ts
```diff
@@ -131,15 +138,19 @@ export function deployAndDisplaySolution(
...
   // Deploy a solution described by the supplied id
   const options: common.IDeploySolutionOptions = {
     jobId: templateSolutionId,
     progressCallback: progressCallback,
     consoleProgress: true,
     storageAuthentication: srcAuthentication,
-    enableItemReuse
+    enableItemReuse,
+    templateDictionary: {
+      params: JSON.parse(customParams)
```
MikeTschudi, line 152: If a solution template is supplied and the Custom Params input box is left alone, "Deploying..." appears on the screen, the app hangs, and the debugger shows "Uncaught SyntaxError: Unexpected end of JSON input".

demos/src/deploy-solution-main.ts
```diff
@@ -107,15 +110,19 @@ export function deploySolution(
...
   // Deploy a solution described by the supplied id
   const options: common.IDeploySolutionOptions = {
     jobId,
     progressCallback: progressCallback,
     consoleProgress: true,
     storageAuthentication: srcAuthentication,
-    enableItemReuse
+    enableItemReuse,
+    templateDictionary: {
+      params: JSON.parse(customParams)
```
MikeTschudi, line 124: If a folder of solution templates is supplied and the Custom Params input box is left alone, "Deploying..." appears on the screen and then is replaced by "{}".

MikeTschudi (changes requested):
The Custom Params box seems as if it should be optional if one doesn't have custom parameters, but this doesn't work: "{}" has to be supplied as a minimum. Is this a requirement? Can we change the web page to something like
```js
const customParams = document.getElementById("customParams").value || "{}";
```
so that the custom parameters have a default value?

jmhauck:
Thanks @MikeTschudi

MikeTschudi (approved):

Issue #674, opened 2021-06-23
Add support for Velocity Real time and Big Data Analytic item types


chris-fox:
Velocity has there own api for reading, creating and deleting analytics. You can see their api documentation here for our localdeployment org:

https://us-iot.arcgis.com/usadvanced00/pipftfmrv9d1divn/iot/api/swagger.html#/BigData_Analytics

This portion of the url will change depending on the organization:

https://us-iot.arcgis.com/usadvanced00/pipftfmrv9d1divn/

You can get this information from the org by requesting the subscriptionInfo on the portal, and it will return something like below if Velocity is enabled where you can get the velocityUrl

https://www.arcgis.com/sharing/rest/portals/self/subscriptionInfo
"orgCapabilities": [{
	"id": "velocity",
	"test": false,
	"level": "Advanced",
	"region": "US",
	"status": "active",
	"endDate": 1632700800000,
	"itemUnits": 0,
	"computeUnits": 0,
	"velocityUrl": "https://us-iot.arcgis.com/usadvanced00/pipftfmrv9d1divn",
	"storageUnits": 0
}
]

We can then use their api to request the json for the real-time and big-data analytics and store that with the template's data. During deployment we can use the api to create a new real-time or big-data analytic using the json defined in the template.

Analytics can point to hosted feature layers in the portal and they are referenced by their item id's. For example here is an example analytic, it has a sources property. I don't feel strongly that we need to add logic to automatically discover these dependencies, if we could use the logic we use for ExB for example where we automatically find item id references that are shared with the same group during creation and replace the item id references with a variable I believe that would work.

	"sources": [{
			"id": "7afde56d-646e-75cc-a53e-4613cc34d359",
			"name": "feature-layer",
			"label": "Snow Routes",
			"schemaTransformation": {..},
			"properties": {
				"feature-layer.portalItemId": "0d6893904c4d4191b5c2312e60e8def7",
				"feature-layer.outSR": "4326",
				"ui": {
					"top": 10,
					"left": 10,
					"width": 240,
					"height": 70
				}
			}
		}

Below are 2 item ids that can be used to test with:

Real Time Analytic: a728f0cf8bdc4fb19749cc1cbad1651b
Big Data Analytic: beeebfe193554397b088e570a47c735a

jmhauck:
@chris-fox what about feeds?

chris-fox:
@jmhauck, let's add support for feeds as well. Thanks.

jmhauck:
[...]

PR #708, opened 2021-08-23
velocity updates

jmhauck:


packages/common/src/velocityHelpers.ts
```diff
@@ -0,0 +1,130 @@
...
+        const orgCapabilities = getProp(subscriptionInfo, "orgCapabilities");
+        /* istanbul ignore else */
+        if (Array.isArray(orgCapabilities)) {
+          orgCapabilities.some(c => {
+            velocityUrl = c.velocityUrl;
+            return velocityUrl;
+          });
+        }
+        // add the base url to the templateDictionary for reuse
+        templateDictionary.velocityUrl = velocityUrl;
```
MikeTschudi, line 58: Do you intend to set `templateDictionary.velocityUrl` to "" in the case where "orgCapabilities" is not returned?

MikeTschudi (approved):

packages/common/src/velocityHelpers.ts
jmhauck: @MikeTschudi good catch...the "" will be ok but I do need to make some changes to properly avoid redundant and bad requests when the org doesn't support velocity.

jmhauck: @MikeTschudi we will prevent users from deploying solutions that have velocity items via an upcoming change to the app but I didn't handle it correctly for users of solution.js outside the app

jmhauck:
#674

Issue #706, opened 2021-08-18
Relates on on some views fail to create due to chunking. 


shoe913:
Some views do not get the correct relationships created because layers required for the view aren't yet created (they are in a subsequent chunk).

1) Deploy Electric Distribution Data Management for AGOL
2) Look at the ElectricDistributionSystem Layer at REST (https://services2.arcgis.com/bg1h7Ix67olZW9Ja/ArcGIS/rest/services/ElectricDistributionSystem_ff294417b03f49a1bf169c6b22ed11c3/FeatureServer)
3) Go to the Poles and Medium Voltage Transformer layers at REST, note that both have Relationships listed
4) Look at the ElectricDistributionSystem_Editing layer at REST (https://services2.arcgis.com/bg1h7Ix67olZW9Ja/arcgis/rest/services/ElectricDistributionSystem_editing_ff294417b03f49a1bf169c6b22ed11c3)
5) Go to the Poles and Medium Voltage Transformer layers at REST, note that Poles still has it relationship listed, but Medium Voltage is missing a relationship.

shoe913:
@chris-fox according to @jmhauck, we may be able to get around this by leveraging the timeout fixes and avoid chunking completely.  

Is there a specific reason we would still need to maintain chunking for layer creation out side potential timeouts?

shoe913:
Verified

chris-fox:
@shoe913, @jmhauck, i was testing deploying services in Enterprise and this additional addToDef to add the relationships was failing on views, causing the deployment to failure. You can repro this using the Sewer Data Management solution in the enterprise certify demo app. I could reproduce this at 10.9 and 10.9.1.

I am curious if this is necessary to do for Enterprise.

shoe913:
@chris-fox , I think we may need to. 

@jmhauck, this is the same issue we were experiencing with the Electric Distribution yesterday on Enterprise.

jmhauck:
@chris-fox could we get some info on how to add relationships to a view after it has been created. 

The situation falls back to the same thing that started this issue. No problem that I see with the source service. It gets all the expected relationships. Typically all of the relationships will be added automatically to the layers in the view based on the relationships that exist in the source service. However, when we submit the addToDef calls in smaller chunks of the layers its possible that not all layers that participate in a given relationship are in the same addToDef call. When this happens the relationship from the source does not get created automatically. 

In online we can simply make an addToDef call after all layers have been added and add the relationships. 

[...]

PR #717, opened 2021-09-15
add relationships for views after layers when layer chunking is used

jmhauck:
#706


MikeTschudi (approved):
Feature services have *so* many weird challenges!

PR #720, opened 2021-09-22
Updated hub.js due to breaking change

MikeTschudi:
Version 8.21.0 introduced a breaking change by redefining its `IHubRequestOptions` interface. This PR fixes our build by replacing that interface with the new `IHubUserRequestOptions` interface.


dbouwman:
@MikeTschudi gah - we thought we made that change in a way that it would not be breaking (hence the 8.21 vs 9.0) but clearly... we were wrong. Sorry for causing this unexpected work.

dbouwman (approved):

Issue #743, opened 2021-10-27
Failing to create feed due to a duplicate name


chris-fox:
We are seeing an issue where a user renames a feed, but the original feed name is still not released. For example they deploy a solution with the name "My Custom Feed". Then through Velocity or the item details you rename the feed to "Hello World Feed". This doesn't actually release the name "My Custom Feed".

As a result when you go to deploy the solution again, we request all the Feed names from the api and they return "Hello World Feed", but not "My Custom Feed". Then when we go to create the feed with the name of "My Custom Feed" it fails with the error like below:

{"status":"error","messages":[{"key":"ITEM_MANAGER__CREATE_FEED_FAILED_DUPLICATE_NAME","args":["My Custom Feed"],"englishMessage":"A published feed with the name AVL Location Only Feed 1 already exists with the organization. Feed names must be unique across the organization. Please use a different name."}]}

Looking in the velocity app, the UI does the same thing, it attempts to create a Feed with that name and then if it fails shows an error to the user. Can we catch this error and if it is this error than try "My Custom Feed 1", and keep incrementing until it succeeds? There doesn't appear to be any other api to determine if the feed name is unique.

chris-fox:
verified

PR #746, opened 2021-10-29
add error checks

jmhauck:
#743


packages/velocity/src/helpers/velocity-helpers.ts
```diff
@@ -172,7 +182,7 @@ export function getTitle(
   authentication: UserSession,
   label: string,
   url: string
-): Promise<string> {
+): Promise<any> {
```
MikeTschudi, line 185: Could the return be typed as something like
```ts
interface IVelocityTitle {
  label: string;
  titles: string[];
}
```
so that we get the benefit of TypeScript's type checking?

MikeTschudi (approved):

Issue #745, opened 2021-10-28
createSolution demo app uses the wrong credentials when creating solution


chris-fox:
When using different org/portal to create a solution in than the org that has the original items the creation will fail if the items are not public, What is happening is the source credentials are used to query the original item or group, but then the destination credentials are used to query the items inside the group or dependencies of the item. 

I discovered this when attempting to create a new solution item in dev from items in production that were not public. 

![image](https://user-images.githubusercontent.com/6796763/139283091-ec887dd6-bcfe-4cc9-ac2e-bc5cfbf622fd.png)

PR #749, opened 2021-11-08
Update auth

jmhauck:
#745


MikeTschudi (approved):
Thank you, John. Good update given how often private sources come up.

Issue #750, opened 2021-11-11
QuickCapture: OIC item's serviceURL property is not correct after deploy a template by connecting to Enterprise portal


xiao8579:
**Describe the bug**
When deploying the Esri template that has the new OIC serviceURL pattern, if a user has been connecting to Enterprise portal from the QuickCapture AGO site. The `serviceURL` has www.arcgis.com as the portal, but it should use the Enterprise portal.

**To Reproduce**
1. Go to https://devext.arcgis.com/apps/quickcapture
2. Sign in with ArcGIS Enterprise https://rpubs19020.ags.esri.com/portal (creator1 | portalaccount1)
3. New template > Windshield Damange Assessment > Use
4. In the deployed OIC item, check its item `data`, the portal url is www.arcgis.com
"ServiceURL": "https://www.arcgis.com/home/item.html?id=02a7c44e3f504a20a4c9f37b6c91e213&sublayer=0"

**Expected behavior**
In the deployed OIC item, check its item `data`, the portal url should be https://rpubs19020.ags.esri.com/portal
"ServiceURL": "https://rpubs19020.ags.esri.com/portal/home/item.html?id=02a7c44e3f504a20a4c9f37b6c91e213&sublayer=0"

xiao8579:
This is a high priority for the QuickCapture team. We worked with the Oriented Imagery team to come up with new URL pattern for the `serviceURL` property for the OIC item, and needs it to work properly for both Online and Enterprise environment. We target to release this with Online 9.4, and all our existing templates created by the Solution need to update too with this new pattern. @MikeTschudi Would you mind taking a look?

chris-fox:
@xiao8579, you can fix this now in your Solution item if you edit the json and replace "https://www.arcgis.com" with {{portalBaseUrl}}. So for example instead of : 

"https://www.arcgis.com/home/item.html?id=02a7c44e3f504a20a4c9f37b6c91e213&sublayer=0"

Update to:

"{{portalBaseUrl}}/home/item.html?id=02a7c44e3f504a20a4c9f37b6c91e213&sublayer=0"

This variable will be replaced when deployed with the portal url. So it will work for Online and Enteprrise.

xiao8579:
@chris-fox Thanks, I will give it a go. 👍

xiao8579:
[...]

PR #751, opened 2021-11-11
add portal base url var and templatize the id

jmhauck:
#750


MikeTschudi (approved):
Thank you for taking care of this, John

Issue #765, opened 2022-01-06
Update logic for Groups source-xxxx tag to use typeKeywords instead


chris-fox:
When we create an item we write a typeKeyword to the item so we know what original item it came from so that it can be reused by solutions that share the same items. For group we have be writing this as a tag, because the typeKeyword wasn't previously supported.

It now is and we can write this typeKeyword to the group instead, which is better because users can't see or modify tags through the UI. Let's update the logic for groups to use typeKeywords instead.

For the searching on reuse we need to search both tags and typeKeywords to support previously deployed versions of the group. We should be able to do this as an Or in the query.

shoe913:
Verified

Issue #761, opened 2021-12-08
URL parameter is added to web maps upon solution item creation


shoe913:
Adding the URL parameter to web maps during item creation prevents them from being downloaded and updated in ArcGIS Pro. 

During creation, we need to ensure that the URL property on for web maps is set to null.

chris-fox:
I would amend and say, if the url is null we should keep it as null, if there is a url in the original web map than we should handle it like we do other urls. I am not sure all the scenarios, I think the idea is we should not be changing it from what it was originally set as.

jmhauck:
@shoe913 could I see an example where this happens? I see the spot in the code where we set this but have not been able to reproduce the issue at this point. I tried with maps that I created or after looking at a few maps from the first couple of solutions I tried after deploying.

shoe913:
@jmhauck I'm still seeing this with the demo app. 

If you deploy the Electric Underground Inspections solution (f6b16af72e1d42cfb371b3d5665df947), the Underground Inspections map is still getting a URL added in the JSON. 

The original content has a value of null for this parameter.

chris-fox:
@shoe913, I don't believe this fix should fix already authored solution items with this url property. We should check that this url property is not set in newly authored solution items via the creation app.

For items in our repro that already have this problem, we should fix in the repo for the next release.

shoe913:
Solution items in the solution item repo have been fixed with this [PR](https://devtopia.esri.com/WebGIS/solution-items/commit/7cb74315220faabc7df7f81ef72f48c4d0792063) 

I used the demo apps to create a new solution item from this group: df635643771a4418810efbdf69ad3272 
The following solution item was created: https://localdeployment.maps.arcgis.com/home/item.html?id=dbfb695ec3a14c9eb1e0b4759266f5fb

Looking at map 1ce51b0c4de84ade86ccc20bdb032258 (Curbside Service Schedules) in the original and then the deployed content, I see the following:
- The URL property is on the original map and then transposed into the solution item and deployed content. 
- For both the deployed and original, the map cannot be opened and updated in ArcGIS Pro. 

I'm not sure how the URL property was added to the original map ( I didn't do it manually) or what it is responsible for. The ArcGIS Pro team has addressed this issue on their end (issues 2893) to ignore the url property. 

However I think our solutions should be standardized in what they have in this parameter. We should either undo- the above PR in the solution items repo to keep the url property filled or, set all new solutions to have the URL property as null in web maps. 

[...]

PR #766, opened 2022-01-10
check group typeKeywords for source references

jmhauck:
#765


MikeTschudi (commented):
Shouldn't these changes include converting a group's tag into a typeKeyword upon creating and deploying a solution for backwards compatibility & automatic updating to the new approach?

jmhauck:
@MikeTschudi The source-id typeKeyword is written on deploy rather than during create. For backwards compatibility groups tags and typeKeywords are checked on deploy to support find exiting items/groups. Only real change for create is allowing a groups preexisting typeKeywords to be retained.

MikeTschudi (approved):

jmhauck:
#761 also went in with this PR

PR #767, opened 2022-01-11
typeKeywords for older groups

jmhauck:
#765

@MikeTschudi one spot I missed. Older group templates won't have the typeKeywords array.

### DIFF
--- a/packages/deployer/src/deploySolutionFromTemplate.ts
+++ b/packages/deployer/src/deploySolutionFromTemplate.ts
@@ -103,23 +103,7 @@ export function deploySolutionFromTemplate(
         }
 
         // update template items with source-itemId type keyword
-        solutionTemplateData.templates = solutionTemplateData.templates.map(
-          (template: any) => {
-            const sourceId: string = "source-" + template.itemId;
-            /* istanbul ignore else */
-            if (template.item) {
-              /* istanbul ignore else */
-              if (template.item!.typeKeywords) {
-                template.item!.typeKeywords!.push(sourceId);
-              } else if (common.getProp(template, "item.type") === "Group") {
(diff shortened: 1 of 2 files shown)


packages/deployer/src/deploySolutionFromTemplate.ts
```diff
@@ -347,6 +335,33 @@ export function deploySolutionFromTemplate(
...
+  templates: common.IItemTemplate[]
+): common.IItemTemplate[] {
+  return templates.map(
+    (template: any) => {
+      const sourceId: string = "source-" + template.itemId;
+      /* istanbul ignore else */
+      if (template.item) {
+        if (template.item!.typeKeywords) {
+          template.item!.typeKeywords!.push(sourceId);
+        } else /* istanbul ignore else */ if (common.getProp(template, "item.type") === "Group") {
```
MikeTschudi, line 354: Is it necessary to test for the group type? This clause could simply handle any item/group that doesn't have `typeKeywords` for any reason.

MikeTschudi (approved):

packages/deployer/src/deploySolutionFromTemplate.ts
jmhauck: @MikeTschudi I wasn't sure if some item types could/would be expected to not have typeKeywords.

jmhauck: Happy to update if you think it's best though

MikeTschudi: The REST API documentation doesn't say what happens when there are no typeKeywords, so this struck me as a convenient way to survive the API not returning an empty list when there are no typeKeywords.

Issue #777, opened 2022-02-03
Add post process of variables for groups


chris-fox:
In some solutions we have circular dependencies that require post processing of variables. We currently handle this for items, but not group details.

In the solution below their is a Hub Site (dc3ad4e557b74f8ca72f09991be009ae) that has a dependency on a group (875df0accee14898b7e10c018ec3590d). That group has a reference to the url of the hub site in its description property ({{dc3ad4e557b74f8ca72f09991be009ae.url}}). The group is created first, and all the references to the group in the hub site are resolved because the hub site is created 2nd. However, the variable in the group is not swizzled because we need to do a 2nd pass on the group to update the url variable of the hub site.

Solution item: 402fe849dfbd4ff3af925277c70c70e4

chris-fox:
Verified

PR #778, opened 2022-02-04
post process groups

jmhauck:
#777

### DIFF
--- a/packages/common/src/restHelpers.ts
+++ b/packages/common/src/restHelpers.ts
@@ -98,7 +98,9 @@ import {
   SearchQueryBuilder,
   setItemAccess,
   shareItemWithGroup,
-  updateItem as portalUpdateItem
+  updateItem as portalUpdateItem,
+  updateGroup as portalUpdateGroup,
+  IUpdateGroupOptions
 } from "@esri/arcgis-rest-portal";
 import { IParams, IRequestOptions, request } from "@esri/arcgis-rest-request";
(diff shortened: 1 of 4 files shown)


packages/common/src/restHelpers.ts
```diff
@@ -1719,6 +1721,36 @@ export function updateItem(
...
 }
 
+/**
+ * Updates a group.
+ *
+ * @param groupInfo The base info of a group; note that this content will be serialized, which doesn't work
+ * for binary content
+ * @param authentication Credentials for request
+ * @param additionalParams Updates that are put under the `params` property, which is not serialized
+ * @return
```
MikeTschudi, line 1731: What is the return, please?  And can we put a real type in instead of "any" for additionalParams and the return?

packages/group/test/group.test.ts
```diff
@@ -1447,5 +1447,62 @@ describe("Module `group`: manages the creation and deployment of groups", () =>
...
+
+      group.postProcess(
+        itemId,
+        "Group",
+        itemInfos,
+        itemTemplate,
+        [itemTemplate],
+        templateDictionary,
+        MOCK_USER_SESSION
+      ).then(() => done.fail, done)
```
MikeTschudi, line 1505: Is there anything that can be tested in addition to successful return from the call?

MikeTschudi (approved):

PR #789, opened 2022-02-23
Bug/fix demo imports

MikeTschudi:
* Reverted demo build module type
* Updated hub.js in demos
* Removed unnecessary tsconfig options (tsconfig defaults for these are better)
* Updated package-locks to match 1.3.1 release


jmhauck:
@MikeTschudi I still don't get why but all looks good when I work from vs code terminal rather than git bash. install, build, test, deploySolution all worked fine

jmhauck (approved):

PR #791, opened 2022-02-23
avoid unnecessary call to findTransformations

jmhauck:


MikeTschudi (approved):
That was quick! Thank you for tracking this down.

PR #792, opened 2022-02-28
Update storage name

jmhauck:


packages/common/test/resources/convert-item-resource-to-storage-resource.test.ts
```diff
@@ -66,3 +67,35 @@ describe("convertItemResourceToStorageResource, template version 1", () => {
     expect(chk.filename).toBe("happy.png");
   });
 });
+
+describe("convertItemResourceToStorageResource, file types", () => {
```
MikeTschudi, line 71: Let's add the undefined case. The path must be handled somewhere because of the 100% coverage, but it seems more complete to handle it at the function level.

MikeTschudi (changes requested):
This is a clear improvement over the way that it was originally handled; thank you.

PR #798, opened 2022-03-22
fix source code links in API Reference

AlexanderBeyn:
The links to source code in API Reference seem to be missing the `packages/` subdirectory.


MikeTschudi (approved):
Thank you very much, Alexander!

Issue #871, opened 2022-09-12
Add layer one at a time in multisource view


chris-fox:
There is a limitation right now in ArcGIS Online that if you attempt to add multiple layers in a single addToDefinition call to a multisource view then every layer after the 1st layer is broken. Instead if you add them one at a time everything works. I believe we added logic in the past that if in a multisource view one layer has a datasource using a different layer in the view we sequenced the order in which we added the layers (#587).

This issue would add to that and just process all multisource views one layer at  a time. Below is an example service that contains 2 layers in a multisource view. With the logic today the view is successfully created using solution.js, but the 2nd layer is broken, you can't query anything from it. Adding them one layer at a time works successfully.

https://localdeployment.maps.arcgis.com/home/item.html?id=52cf645e0ef64456a1e3135c27a7318c

chris-fox:
verified

PR #873, opened 2022-09-12
Update for ms views

jmhauck:
#871


MikeTschudi (approved):
Is there a way for the tests to catch that the layers are being added one at a time?  I don't see how the existing tests are helping us.

jmhauck:
@Mike I updated the test to hopefully be more clear. This function will check if its a multi-service view and when it is we should create a new array for the current layer or table to be added to so it will be processed separately from any other layers in the multi-service view.

Issue #1047, opened 2023-03-22
View referencing fields in the source that based on statistics are being removed


chris-fox:
When deploying to enterprise if you have a view that is referencing fields in a source view that were generated from statistics those fields are dropped in the addToDefinition for the layer.

1. Deploy this item to Enterprise: f6f08c93252a41ba8d45538012ab2ded
2. The resulting Gravesites_labels view is missing the burialcount, fullname1, and fullname2 fields

The solution has 3 feature layers (1 feature service and 2 views). The Gravesite_burials view has a table id 1 that has some fields defined as statistics, these fields don't exist in the source service, they are dynamically generated and have a property statisticType.
```
 "sourceLayerFields" : [
          {
            "name" : "parentglobalid", 
            "alias" : "Parent Global ID", 
            "source" : "parentglobalid"
          }, 
          {
            "name" : "burialcount", 
            "alias" : "Burial Count", 
            "source" : "parentglobalid", 
            "statisticType" : "COUNT"
          }, 
          {
            "name" : "fullname1", 
            "alias" : "Full Name 1", 
            "source" : "fullname", 
            "statisticType" : "MIN"
          }, 
          {
            "name" : "fullname2", 
            "alias" : "Full Name 2", 
            "source" : "fullname", 
            "statisticType" : "MAX"
          }
        ]
```

The Gravesites_labels view has a layer that on the right-side of the join uses the table above as a source and references those statistic fields of burialcount, fullname1, and fullname2.
```
"relatedTables" : [
          {
            "name" : "Cemeteries_StatsTable", 
            "sourceServiceName" : "Gravesites_burials_dd6da282de23402ba4b82e4ca0370536", 
            "sourceLayerId" : 1, 
            "sourceId" : 98822, 
            "sourceLayerFields" : [
              {
                "name" : "parentglobalid", 
                "alias" : "Parent Global ID", 
                "source" : "parentglobalid"
              }, 
              {
                "name" : "burialcount", 
                "alias" : "Burial Count", 
                "source" : "burialcount"
              }, 
              {
                "name" : "fullname1", 
                "alias" : "Full Name 1", 
                "source" : "fullname1"
              }, 
              {
                "name" : "fullname2", 
                "alias" : "Full Name 2", 
                "source" : "fullname2"
              }
            ], 
            "type" : "LEFT", 
            "parentKeyFields" : ["globalid"], 
            "keyFields" : ["parentglobalid"], 
            "operator" : "EQUALS"
          }
        ]
[...]

PR #1051, opened 2023-03-23
Field issue

jmhauck:
#1047


packages/common/src/featureServiceHelpers.ts
```diff
@@ -1212,6 +1223,58 @@ export function _getFieldNames(
...
+  const expressions = table.sourceLayerFields.reduce((prev,cur) => {
+    /* istanbul ignore else */
+    if (cur.source) {
+      const hasExpression = cur.expression ? true : false;
+      if (Object.keys(sourceFieldInfos).indexOf(cur.source) > -1) {
+        const source = sourceFieldInfos[cur.source];
+        source.hasExpression = source.hasExpression || hasExpression;
+        source.names.push(cur.name);
+        /* istanbul ignore else */
+        if (source.hasExpression || hasExpression) {
```
MikeTschudi, line 1251: This line seems unnecessary because `source.hasExpression` has already been adjusted on line 1248

packages/common/test/featureServiceHelpers.test.ts
```diff
@@ -6745,6 +6746,64 @@ describe("Module `featureServiceHelpers`: utility functions for feature-service
...
+            "source": "population"
+          }
+        ],
+        "groupBy": "parentglobalid",
+        "materialized": false
+      };
+
+      const actual = _getDynamicFieldNames(table);
+
+      const expected = ["burialcount", "fullname1", "fullname2", "population", "area"];
```
MikeTschudi, line 6802: `area` looks like a source field, as opposed to `density`, which has an expression. What am I missing, please?

MikeTschudi (approved):

jmhauck:
@MikeTschudi The expression check came up as a possible thing for us to handle but it's not represented in the issue so I pulled for now. Will add back if necessary after chatting with Chris.

MikeTschudi (approved):

Issue #1048, opened 2023-03-22
View fails to deploy if a layer has a source to a different layer or table in the same service


chris-fox:
Deploy the following item, it will fail on the addToDefinition for the view.

e70e8d56824145b1a3c8356f51bf56f8

In this solution there are 2 feature layers (1 source service and 1 view). The view has 1 layer (id 1) and 1 table (id 0). Table id 0 is pointing to the source service. Layer id 1 is a join view between a layer in the source service and the table id 0 in the view service. The issue is when creating the view it is attempting to add Layer id 1 first before Table id 0 has been added and this causes it to fail.

In these scenarios when a layer or table references a layer or table in the same feature layer or view we need to process them in order to ensure all the sources are added first.

jmhauck:
@chris-fox can the id be used for sorting or just circumstance that we want id 0 first in this case?

jmhauck:
And thanks for the info on using IDs...I had added a topological sort that looked through what was related to what and defined the order that way...similar to what we do for defining the overall build order but IDs will be easier.

chris-fox:
The original item can no be successfully deployed, but I am encountering a similar issue in a different solution item that is still failing because the layers/tables are being added in the wrong order.

2f3b031686c347f9be27cc84ee2a8739

chris-fox:
Verified

PR #1103, opened 2023-05-10
Id check

jmhauck:
#1048

### DIFF
--- a/packages/common/src/featureServiceHelpers.ts
+++ b/packages/common/src/featureServiceHelpers.ts
@@ -1100,9 +1100,25 @@ export function _isSelfReferential(
   layersAndTables: any[]
 ): boolean {
   const names = layersAndTables.map(l => l.item.name);
+  const srcTables = {};
   return layersAndTables.some(l => {
-    const relatedTables = l.item.adminLayerInfo?.viewLayerDefinition?.table?.relatedTables || [];
-    return relatedTables.some(r => names.indexOf(r.name) > -1);
+    const table = l.item.adminLayerInfo?.viewLayerDefinition?.table;
+    if (table) {
+      const name = table.sourceServiceName;
+      const id = table.sourceLayerId;
+      if (name && id > -1) {
+        if (Object.keys(srcTables).indexOf(name) > -1) {
+          if (srcTables[name].indexOf(id) > -1) {
+            return true;
+          } else {
+            srcTables[name].push(id);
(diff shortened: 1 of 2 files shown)


packages/common/src/featureServiceHelpers.ts
```diff
@@ -1100,9 +1100,25 @@ export function _isSelfReferential(
...
+          if (srcTables[name].indexOf(id) > -1) {
+            return true;
+          } else {
+            srcTables[name].push(id);
+          }
+        } else {
+          srcTables[name] = [id]
+        }
+      }
+      return (table.relatedTables || []).some(r => names.indexOf(r.name) > -1);
```
MikeTschudi, line 1120: Same question here. `some` returns truthy or falsy.  If `some` were to return the falsy value "0", this condition would still return true.

packages/common/src/featureServiceHelpers.ts
```diff
@@ -1100,9 +1100,25 @@ export function _isSelfReferential(
...
   const names = layersAndTables.map(l => l.item.name);
+  const srcTables = {};
   return layersAndTables.some(l => {
-    const relatedTables = l.item.adminLayerInfo?.viewLayerDefinition?.table?.relatedTables || [];
-    return relatedTables.some(r => names.indexOf(r.name) > -1);
+    const table = l.item.adminLayerInfo?.viewLayerDefinition?.table;
+    if (table) {
+      const name = table.sourceServiceName;
+      const id = table.sourceLayerId;
+      if (name && id > -1) {
```
MikeTschudi, line 1109: Why not `if(name && id) {`?

MikeTschudi (changes requested):

packages/common/src/featureServiceHelpers.ts
jmhauck: indexOf(r.name) > -1 will return boolean

packages/common/src/featureServiceHelpers.ts
MikeTschudi: Thank you--I got operator precedence wrong

packages/common/src/featureServiceHelpers.ts
MikeTschudi: Thank you--I'd misread the line

MikeTschudi (approved):

Issue #1151, opened 2023-06-27
Duplicate but bogus resources listed when creating a Solution out of a Quick Capture Project


MikeTschudi:
When creating a Solution out of a QC project, solution.js pre-fetches its resources, but then runs the project through the same resource fetching that it does for other item types. The initial set is not in the correct format--they're files--and create a Solution that can't be deployed:
![image](https://github.com/Esri/solution.js/assets/2125181/a71ea99a-1143-4ef8-8792-e22f7dbe88f1)

chris-fox:
Verified

PR #1152, opened 2023-06-28
Fix handling of QuickCapture resources

MikeTschudi:
Also removed reporting result of sanitizing

#1151


jmhauck:
@MikeTschudi this addresses the question in Foxs email because the qc.project.json is already stored as a resource by the time we get here?

MikeTschudi:
Yes--no change in handling of qc.project.json.

image

jmhauck (approved):

Issue #1153, opened 2023-06-28
Use Quick Capture Project qc.project.json file rather than `data` for config


MikeTschudi:
Older Quick Capture Projects had their configuration in the item's `data` as well as in the qc.project.json file; newer projects only use qc.project.json.  The difference between the two seems to only be that the contents of the qc.project.json file are wrapped when in `data`:
```json
{
	"application": {
            _qc.project.json contents_
	},
	"name": "qc.project.json"
}
```

solution.js only handles the older format, which means that it only swizzles the contents of `data`.  It copies whatever is (or isn't) in `data` and maintains the qc.project.json file as a resource.

With new format files, the swizzling needs to be done in the qc.project.json file because `data` is empty.

As a hotfix, we will copy qc.project.json into `data` with the appropriate wrapper.  Later, we will change solution.js to handle both formats separately.

chris-fox:
@MikeTschudi, I would like to continue to have it behave like it did previously. When we templatize it, it is helpful for the contents of the gc.project.json to be in the data property. This is because we can use tools like arcgis assistant to do a find/replace all for things across all the templates. If this config is only stored in the item resources it makes it a little more difficult.

I think what we could fix is during deployment we don’t need to write this information to the item data, instead just swizzle it and upload it as the gc.project.json. This change would not break any existing templates as we are not changing the way the template is written.

MikeTschudi:
@chris-fox, my only concern with making the hotfix permanent is whether or not the QC apps will stop supporting backwards compatibility for the `data` section, such as for new features.  At this time, `data` and qc.project.json are essentially identical, but that can't be guaranteed.

@shoe913 noticed that the six release solutions that use QC all have info in their data.json file; it's only with new QC projects that `data` is not used.

By copying qc.project.json into `data` while templatizing, I didn't have to make any changes to deployment.  I think that we haven't needed to templatize qc.project.json for our solutions because of backwards-compatibility support.

chris-fox:
@MikeTschudi, I don't believe QC has made any changes to their item, will try to clarify what I am hoping to continue to see in the behavior of this item type. To my knowledge QC has never used the item data property. All of the properties that define the quick capture project have always been stored in the qc.project.json item resource, this includes the feature services that need to be discovered and templatized.

[...]

PR #1165, opened 2023-07-14
Fix handling of QuickCapture projects

MikeTschudi:
[#1153](https://github.com/Esri/solution.js/issues/1153#issuecomment-1625979173)

Coveralls is not reporting, therefore:
image


MikeTschudi:
Cycled close/open to resync coveralls

PR #1171, opened 2023-07-17
Fix handling of QuickCapture projects

MikeTschudi:
#1153


MikeTschudi:
Cycle closed/opened to sync coveralls

Issue #875, opened 2022-09-16
Add support for Workflow Manager items


chris-fox:
Additional details to come in the future, but Workflow Manager is currently available in Enterprise and is planned to come to Online next year.

chris-fox:
@MikeTschudi, Looking through the api, I think there might be an easier approach to this. There is an [exportWorkflowConfiguration](https://developers.arcgis.com/workflow-manager/api-reference/rest/export-workflow-configuration/) and [importWorkflowConfiguration](https://developers.arcgis.com/workflow-manager/api-reference/rest/import-workflow-configuration/). 

The export creates a .wmc file and will contain all the diagrams, jobTemplates and several other properties configured in the workflow. This could be a much simpler approach than having to handle every property individually. What is not clear to me is what kind of file is the .wmc file. Is it a json file? We need to be able to templatize properties in the diagrams so if it is not a file type that we can look for itemId references than this approach may not work for us.

There is no UI to get a wmc file in Workflow Manager and I have tried to use the python api to get it but I am having trouble authenticating. If you could get a sample .wmc file form the example workflow referenced in the comment above I would like to take a look to see if it might be a viable approach.

MikeTschudi:
Sample .wmc file; it's a zip with a renamed extension.  Because GitHub doesn't support ".wmc", the file is attached with the ".zip" extension.
[workflow_configuration_1691603968445.zip](https://github.com/Esri/solution.js/files/12305626/workflow_configuration_1691603968445.zip)

MikeTschudi:
Within the .wmc file, there appears to be just three files that need to be templatized:
* diagrams.json                       
* jobExtPropertyTableDefinitions.json 
* roleGroups.json

chris-fox:
@MikeTschudi, would it hurt to just templatize all the files. If there ends up being nothing in the file to templatize than it shouldn't hurt anything, but I guess my concern is I don't know enough yet to say which files might have other item references. Also if things change over time we are more likely to just handle the additional cases we are not aware of yet.

chris-fox:
@MikeTschudi, I have been testing this today and have found workflow manager items are failing to deploy. I have been testing with this Workflow item:

https://localdeployment.maps.arcgis.com/home/item.html?id=661e0e38d4ad4a8f97731b22cb6c95dd

It creates successfully, but when I deploy it fails. Looking at the network traffic I am finding it is failing on the import command with the following error:
```
{
    "error": {
[...]

PR #1367, opened 2024-01-31
Add workflow package placeholder

MikeTschudi:
This is a first step for implementing #875. The monorepo setup makes it tricky to reference a new package, so this PR introduces the workflow package so that we can publish a version. Once published, the package should be referenceable by other packages.

Includes PRs #1365 and #1366.


packages/common/src/workflowHelpers.ts
```diff
@@ -0,0 +1,91 @@
...
+ */
+
+import * as zipUtils from "./zip-utils";
+
+// ------------------------------------------------------------------------------------------------------------------ //
+
+export function compressWorkflowIntoZipFile(
+  //workflow: any
+): Promise<File> {
+
```
jmhauck, line 30: @MikeTschudi this is coming later?

packages/workflow/src/workflow.ts
```diff
@@ -0,0 +1,57 @@
...
+ * @param srcAuthentication Credentials for requests to source items
+ * @returns A promise that will resolve when the template has been created
+ */
+export function convertItemToTemplate(
+  /*solutionItemId: string,
+  itemInfo: any,
+  destAuthentication: common.UserSession,
+  srcAuthentication: common.UserSession*/
+): Promise<common.IItemTemplate> {
+  return Promise.resolve(commonTest.getItemTemplate("Workflow"));
```
jmhauck, line 43: @MikeTschudi actual implementation is coming later?

MikeTschudi: Yes. Once I'm past the new-package problem, I'll have a better sense of how to manage the config files zip via this module.

PR #1370, opened 2024-02-01
Remove references to Coveralls

MikeTschudi:
Because we no longer use Coveralls (we use Codecov), these values aren't being updated. Codecov does not appear to do package-level coverage.


MikeTschudi:
Coveralls is not working right now.
image

jmhauck (approved):

Issue #1368, opened 2024-01-31
Join View fails to deploy if source services have different spatial references


chris-fox:
Can be reproduced by deploying this solution item:

https://localdeployment.maps.arcgis.com/home/item.html?id=111b98baa09742fc97651ae53f44dbc5

The issue is the 2 source services have different spatial references. In the join service layer id 0 is a table and it has a single source pointing to the service with a spatial reference of 4326. Layer id 1 is a layer and is a multi-source service where the left side of the join is coming from the service with a spatial reference of 102100.

So what appears to be happening is we are pulling the spatial reference from the source of layer id 0. I believe we implemented this logic to account for how users can change the spatial reference on deployment and the need to keep views in sync with the source service spatial reference.

However, this is a unique case as the layer id 0 is a table and has no geometry so spatial reference of the source can be ignored. I believe we should update this logic to look at the spatial reference of the source service for the first layer (w/ geometry) in the service. So in this particular case we would take the spatial reference from the source service on the left side of the join for layer id 1.

PR #1417, opened 2024-03-13
Spatial reference issue

jmhauck:
#1368

### DIFF
--- a/packages/common/src/featureServiceHelpers.ts
+++ b/packages/common/src/featureServiceHelpers.ts
@@ -1497,8 +1497,41 @@ export function validateSpatialReferenceAndExtent(
 ): void {
   /* istanbul ignore else */
   if (getProp(serviceInfo, "service.isView")) {
+    const layersAndTables = [
+      ...serviceInfo?.layers || [],
+      ...serviceInfo?.tables || []
+    ];
+
     let sourceSR: any;
     let sourceExt: any;
+
+    // first pass ensure we have a geometry type before getting the spatial reference or extent
+    // issue: #1368
+    itemTemplate.dependencies.some(id => {
+      const source: any = templateDictionary[id];
+
+      let hasGeom = layersAndTables.some(layerOrTable => {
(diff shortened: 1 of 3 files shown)


packages/common/src/featureServiceHelpers.ts
```diff
@@ -1539,6 +1541,58 @@ export function validateSpatialReferenceAndExtent(
...
+ * @param templateDictionary Hash mapping Solution source id to id of its clone (and name & URL for feature service)
+ * @param validateGeom When true the source must contain a geometryType for values to be returned
+ * @private
+ */
+export function _getSourceSpatialReferenceAndExtent(
+  serviceInfo: any,
+  itemTemplate: IItemTemplate,
+  templateDictionary: any,
+  validateGeom: boolean
+): any {
```
MikeTschudi, line 1559: Please consider a local interface definition that can be used to type the result.

packages/common/src/featureServiceHelpers.ts
```diff
@@ -1539,6 +1541,58 @@ export function validateSpatialReferenceAndExtent(
...
 
+/**
+ * Get the spatial reference from a views source.
+ * Optionally ensure that the source has a valid geometry type prior to using its values.
+ *
+ * @param serviceInfo Basic service information
+ * @param itemTemplate The current template to process
+ * @param templateDictionary Hash mapping Solution source id to id of its clone (and name & URL for feature service)
+ * @param validateGeom When true the source must contain a geometryType for values to be returned
+ * @private
```
MikeTschudi, line 1552: Please add return doc.
Is it significant to doc that (after passing the has-geometry test if `validateGeom` is true) what is returned is the first view that matches having _both_ a SR and extent?

MikeTschudi (dismissed):

packages/common/src/featureServiceHelpers.ts
jmhauck: Just checking for both in the [].some to know if we can bail early.

PR #1448, opened 2024-04-17
Revise workflow handling

MikeTschudi:
Fix the discovery and handling of the URL to the workflow manager.

Rename variables as needed to better match their purpose after having learned more about workflow manager.

#875

(There are a gazillion commits because I had to isolate a bug occurring only during testing.)


MikeTschudi:
codecov appears to be struck
image

jmhauck (approved):

PR #1496, opened 2024-09-06
fix id checking

jmhauck:


packages/creator/src/createItemTemplate.ts
```diff
@@ -550,30 +550,33 @@ export function _templatizeResources(
...
                   if (isValid) {
                     const id: string = idLookup[i];
-                    // templatize the itemId--but only once per unique id
-                    const regEx = new RegExp(id, "gm");
-                    dataString = dataString.replace(regEx, "{{" + id + ".itemId}}");
-
-                    // update the dependencies
-                    if (itemTemplate.dependencies.indexOf(id) === -1) {
-                      itemTemplate.dependencies.push(id);
+                    if (verifiedIds.indexOf(id) < 0 && id) {
```
MikeTschudi, line 570: * Can `id` be zero?
* How about checking it before looking it up in verifiedIds?

MikeTschudi (approved):

chris-fox:
@jmhauck, not sure if this is what you are doing, but I have been thinking it might be safest and easiest to just remove the logic to auto-discover item ids in python notebooks. Instead if we could just do a post process of a notebook after all the items have been discovered and swizzle any item ids that are present in the notebook and add those items as dependencies of the notebook? 

This is similar approach we take in other item types and works well. People just need to include the items in the group to ensure they are packaged and swizzled with the notebook. The other reason I think this would be good is sometimes people might want to reference an external resource in a notebook, like living atlas content that they don't want to package with the notebook.

jmhauck:
@chris-fox this change is for https://devtopia.esri.com/WebGIS/solution-deployment-apps/issues/172#issuecomment-4990734 

Then for Notebooks...we are currently getting its data and searching for item ids. I will switch it to avoid this. Would I need to check all other item ids from the Solution against the notebooks data section then?

chris-fox:
@jmhauck, yes in the post process, we should search the notebook's data against all other item and group ids from the Solution. In addition, if we could also search for feature service urls from the solution in the notebook's data.

jmhauck:
https://github.com/Esri/solution.js/pull/1499

PR #1501, opened 2024-09-13
Gp service updates

jmhauck:
https://devtopia.esri.com/WebGIS/solution-deployment-apps/issues/100

Still need to investigate one GPServer url that isn't found in the exb app.


previnWong (approved):
LGTM

PR #1520, opened 2025-01-22
Re-templatize a Deployed Solution

previnWong:
Hey guys,

This is more a review right now than a merge.  Can you see if I forgot any gotchas in this work.  Idea is that you can re-templatize a solution you deployed. And below is my assumptions

0) I check if the item is Solution Item and if it has been Deployed
1) I read the item ids from the Solution Item
2) I check if those items still exists and only keep those.
3) I check for new things in the folder that were not in the Solution item and add them to the item id list.
4) then call create Solution on the item id list.

Also, in the Create JS Demo, I removed (hide) the destination credentials.  It causes issues when destination is not the same as source for some items.  Since Creation App only has one login, kept it the same way.

https://devtopia.esri.com/WebGIS/solution-deployment-apps/issues/268


MikeTschudi (commented):
- Perhaps it would be better to create a new demo, e.g., "recreateSolution" rather than modify "createSolution"?
- Is the speed for checking for existing items acceptable?  For speed, you could push the promise for each item or group fetch into an array, and then use Promise.allSettled() to wait for all of the results.

jmhauck:
The general approach sounds good to me.

previnWong:
Ok, let me make a different demo for re-deployed solutions.

I'll try adding a promises.all to the logic.

MikeTschudi:
[`Promise.allSettled`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/allSettled) is what you'd want here, not [`Promise.all`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/all):  the latter "rejects when any of the input's promises rejects", whereas the former waits until all promises have been resolved, whether successfully or with a failure.  One then can go through the list of results filtering out the failures by keeping only ones with a `status` of "fulfilled".

For example,
```ts
const itemFetchPromises = solutionData.templates.map((template: any) => {
  return common.getItemBase(template.itemId as string, userSession);
});
const itemFetches = await Promise.allSettled(itemFetchPromises);
const items = itemFetches
  .filter((item: any) => item.status === "fulfilled")
  .map((item: any) => item.value);
```

previnWong:
@MikeTschudi @jmhauck can you guys review this for merging this time.  Added a new demo so there's alot of files around that making it seem a bigger PR than it is.

packages/creator/src/creator.ts
```diff
@@ -348,3 +378,75 @@ export function _getDeploymentProperty(desiredTagPrefix: string, tags: string[])
     return null;
   }
 }
+
+export async function _updateCreateOptionForReDeployedTemplate(
```
MikeTschudi, line 391: Would you please add function documentation?

MikeTschudi (changes requested):

packages/creator/src/creator.ts
previnWong: right! thanks for the reminder.

jmhauck (dismissed):

packages/creator/src/creator.ts
previnWong: added function doc.

MikeTschudi (approved):

PR #1532, opened 2025-04-07
Patch getDomainCredentials to handle URLs with mismatching cases

MikeTschudi:
It is possible for an Enterprise system to be configured where the casing of domain names in a list of authorized domains doesn't match the URL being requested. Because domain names are case insensitive, this PR changes getDomainCredentials to perform a case-insensitive comparison.


chris-fox:
Just reading through this, is there a reason the demos are being updated to version 6.x? I would like to keep develop at 5.8.x until we are ready to merge in all the changes for the rest-js changes. We need a stable 5.8.x version to go out for enterprise 11.5.

MikeTschudi:
Not a good reason, but it's the result of the attempts to get past the npm publishing failures.  We'd publish and get 11 successes and four failures.  Can't go backwards with npm versions.  I jumped us to 6.x to see if npm was having trouble with the fact that we are using "6.0.x-alpha.0" for the rest-4 work, but that didn't help.

PR #1541, opened 2025-04-25
Republished because of errors in publishing

MikeTschudi:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@esri/solution.js",
-  "version": "5.8.9",
+  "version": "5.8.25",
```
previnWong: any reason why this one is 5.8.25 and the others are 5.8.26?

previnWong (commented):
there was one odd 5.8.25 version in there.

package.json
MikeTschudi: Looks like a bug in the publishing tool or batch file.  I'll update it manually this time.  Thanks for the catch.

previnWong (approved):

PR #1546, opened 2025-05-27
switch subgroups parameter to true

previnWong:
switch subgroups to true in the recreate solution case to handle embedded groups option in demo app.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 ## Version 6 marks the use of ArcGIS Rest JS 4x within Solution JS. Previous versions leveraged ArcGIS Rest JS 3x.
 
 ## RELEASES
-**v6.3.0.next.20250529**
+**v6.3.0.next.***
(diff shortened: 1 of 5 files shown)


jmhauck (dismissed):

jmhauck:
Tests should be updated to stay at 100% coverage before you merge

previnWong:
We were short 100% doing the migration to 4x.  We were going to make an separate issue to address the missing coverage.  I'll try to address them.

previnWong:
@jmhauck brought it up to 100% code coverage.  Removed some uncovered code since a unit test script had no way to create a parameter in the function, so that code logic is never truthsy (restHelper line 700)

There are 2 fixes included.

1) The change to true on the parameter creator file to accommodate the embedded group checkbox in the recreation solution demo app.

2) Added a data type check on AddItemData since Rest JS changed the parameter from data to file or text.  So added a check for instance of file, otherwise text.

jmhauck (dismissed):

jmhauck (approved):

Issue #1240, opened 2023-09-28
Recycle Bin and Deleting Solutions


MikeTschudi:
From @shoe913: 
>The Online R03 release will include the new Recycle Bin capability. There will be 2 modes for deleting content: "Move to Recycle Bin" (default 14 days), and "Permanently Delete".
>
>Not all item types are currently supported for the Recycle Bin . For example Hub Content and Workflows are not supported will be permanently deleted every time.
>
>In order to have a consistent experience with all items when deleting a solution, I believe we should use the "Permanently Delete" option when deleting a solution.
>
>Do do this, we will need to add the property "permanentDelete: true" to the payload for each delete call.

Apply to deleteSolutions demo.

dom-jones-1:
Hi @MikeTschudi, in QuickCapture we would like to use `deleteSolution()` to enable the user to delete all files associated with their QC project, but want to give them the option to send it to the recycle bin or permanently delete.

Is it possible to add a flag to `IDeleteSolutionOptions` (or somewhere else) and to pass this parameter through to `removeItem()`?

chris-fox:
@dom-jones-1, as described in the issue there are some item types that don't support the recycle bin and are always permanently deleted. For example, you get the confirmation window below when you attempt to delete many items in the home app and some do not support the recycle bin:

![Image](https://github.com/user-attachments/assets/71e05ee7-27cf-4457-8e19-931799936a0c)

In addition, an org may have not enabled the recycle bin and in which case the items will all be permanently deleted. 

The REST api seems to handle both cases fine without any additional work from our side, but my questions are:

- If we add this flag, are you ok with depending on the solution and the items included that some item types may be permanently deleted regardless of it you ask that they be sent to the recycle bin or not, because they don't support the recycle bin?
- Do you expect that we also provide some api to return which items can be recycled and which ones can't?

dom-jones-1:
Hi @chris-fox, thanks for your response. We are aware that the recycle bin needs to be enabled at the org level, and we already currently handle that check, so that won't be an issue. Looking through the list of unsupported file types, we think it will be unlikely that a QuickCapture project will include any of these, but to be on the safe side we would like to be able to know if there are un-recyclable items before committing to deleting them.

To answer your points:

1. We are okay with un-recyclable items being permanently deleted, however
2. An API to check this beforehand so we can warn the users there are unsupported file types in their solution would be useful, but we could probably work around this if it wasn't feasible.

chris-fox:
[...]

PR #1550, opened 2025-06-11
Add flag to solution deletion to allow sending items to recycling bin

MikeTschudi:
#1240

Also added flag to deleteSolution demo UI


previnWong (approved):

previnWong:
in deploy app when deleting a solution, do we need to add a checkbox for permanent as well? Or just let it go to recycle bin by default?

chris-fox:
@previnWong, @MikeTschudi, in the Deploy app we want to keep the same behavior as today where everything is delete permanent. This is because not all item types are recyclable and it would be a weird user experiences if some of the items in a solution were recycled but not others. This fix is just for other consumers of solution.js to leverage as they need.

PR #1552, opened 2025-06-17
Recycle solution item when that option is selected in the deleteSolution demo

MikeTschudi:
https://github.com/Esri/solution.js/issues/1240#issuecomment-2980714592

### DIFF
--- a/packages/common/src/deleteHelpers/deleteSolutionItem.ts
+++ b/packages/common/src/deleteHelpers/deleteSolutionItem.ts
@@ -20,7 +20,7 @@
  * @module deleteSolutionItem
  */
 
-import { IDeleteSolutionOptions, IStatusResponse } from "../interfaces";
+import { IStatusResponse } from "../interfaces";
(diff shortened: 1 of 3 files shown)


packages/common/src/deleteHelpers/deleteSolutionItem.ts
```diff
@@ -32,17 +32,23 @@ import { removeItem } from "../restHelpers";
...
  * @param authentication Credentials for the request
+ * @param permanentDelete If true (the default), the item is permanently deleted; if false and the item type
+ * supports the recycle bin, the item will be put into the recycle bin
  * @returns Promise that will resolve with the status of deleting the item
  */
-export function deleteSolutionItem(solutionItemId: string, authentication: UserSession): Promise<IStatusResponse> {
+export function deleteSolutionItem(
+  solutionItemId: string,
+  authentication: UserSession,
+  permanentDelete: boolean,
```
previnWong: should we make this a optional parameter  permanentDelete?: boolean  for backward compatibility?

previnWong (approved):

PR #1554, opened 2025-08-06
addcancelcheck to deploysolution

previnWong:
Added abortController as an optional property on IDeploySolutionOptions interface.

Added abortController function and checks in various sports in the deployment workflow.  I tried to put checks, before and after async callbacks functions. Since the state of abort can change at any time.
Throws an error or rejects, causing the deploy to fall into the Catch clause of deploySolution, and rollback and delete the solution.

Updated unit tests

### DIFF
--- a/packages/deployer/src/deployer.ts
+++ b/packages/deployer/src/deployer.ts
@@ -26,8 +26,6 @@ import { deploySolutionFromTemplate } from "./deploySolutionFromTemplate";
 import { getSolutionTemplateItem, isSolutionTemplateItem, updateDeployOptions } from "./deployerUtils";
 import { IModel } from "@esri/hub-common";
 
-let abortSignal: boolean = false;
-
 /**
  * Deploy a Solution
  *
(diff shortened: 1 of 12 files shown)


packages/deployer/src/deployer.ts
```diff
@@ -103,6 +120,28 @@ export async function deploySolution(
...
     });
 }
+
+export function deployCatchHandler(ex: any, authentication: common.UserSession) {
+  const progressFcn = function () {
+    // Create base progress HTML
+    const html = "Deleting from Deployer";
+
+    // Get the output container
+    const outputEl = document.getElementById("output");
```
MikeTschudi: We shouldn't have html at this level--that's an app purpose. The message and/or code should be thrown for the app to catch and to handle.

MikeTschudi (changes requested):

packages/deployer/src/deployer.ts
previnWong: That code was existing in there in line 122 before (Catch block).  I just moved it to a separate function to help with the unit test to test it easier.
But looking now, yes, it is odd, since there output dom might not even exists.

Should I remove it?

MikeTschudi: Yes, please do remove it or replace it with an exception.

It might be even more interesting if there is an app using the library that has a dom element with the id "output"!

previnWong: hehehe, that block caused me so much Unit Test issues.  So deep in the weeds, didn't think higher picture.

previnWong:
@MikeTschudi changes to remove the dom element check and cleaning up unit test for the removal have been pushed. Please re-review.

MikeTschudi (approved):
Thank you, Previn

PR #1555, opened 2025-08-18
Handle case of no failed, but user cancelled

previnWong:
Handle case of no failed templates, but user cancelled before solution is created/updated.  (for example when a solution has only 1 item template).

Have to handle the delete right in the same function, otherwise the error bubblingup fails because deploySolution (higher up) needs typeKeyword "deployed" to exist t work which at the point of cancel, it does not.


previnWong:
hold on review, have a failing test.

MikeTschudi (dismissed):

jmhauck (approved):

PR #1567, opened 2026-02-11
Add solutionOwner to request options

previnWong:
ArcGIS Rest JS has a function to check owner.  It is done by passing in the owner property in request/delete/update Options. If not it uses the logged on user.  Right now in SolutionJS, we don't pass owner property.  This change adds a solutionOwner prop to some functions and alters the interface with solutionOwner in other places.  So that the URL to request a REST operation does it on the right url path.


jmhauck (dismissed):

jmhauck:
oops...overlooked that this was a draft. Happy to review again when ready.

cosbyr (approved):

PR #1586, opened 2026-05-20
If both "AIAssistant" and "ArcGIS Solution" typeKeywords are present, remove "ArcGIS Solution"

jmhauck:


chris-fox:
@jmhauck, sorry for the confusion on this one. I don't think we need to do this in solution.js, i am just going to remove that typeKeyword on the template itself. I just didn't want to do it yet, because it breaks my demo with the older version of the app.

